PR #66, opened 2026-03-28
Fix Linux x86-64 simulator build compatibility

RegOrton:
## Problem

The TouchGFX simulator could not be built on x86-64 Linux (Ubuntu 24.04, GCC 13). Seven issues were found and fixed:

### `Libs/Header/SDK/Simulator/OS/OS.hpp`
`OS::Queue<T,N>` uses `std::array<T,N>` but `#include <array>` was inside the `#ifdef _WIN32` block only. On Linux, `std::array` had only a forward declaration (pulled in transitively via `<tuple>`), causing an "incomplete type" error at template instantiation. Added `#include <array>` to the `#else` branch.

### `Libs/Header/SDK/Simulator/Kernel/Mock/App.hpp`
`#include <windows.h>` was unconditional at the top of the file. Nothing in the class body uses any Windows-specific type or function. Guarded with `#ifdef _WIN32`.

### `Libs/Header/SDK/Simulator/Kernel/Mock/Logger.hpp`
`GetTickCount64()` is a Win32 API. Added a portable inline wrapper using `clock_gettime(CLOCK_MONOTONIC)` under `#ifndef _WIN32`.

### `Libs/Source/Simulator/Kernel/Mock/System.cpp`
`GetTickCount64()` and `Sleep()` are Win32-only. Added local inline wrappers (`clock_gettime` / `usleep`) under `#ifndef _WIN32`.

### `Libs/Header/SDK/Messages/MessageBase.hpp`
`static_assert(sizeof(MessageBase) == 32)` fails on x86-64 because `void* mCompletionSemaphore` and the vtable pointer are each 8 bytes (not 4), making the struct 40 bytes. The assertion is correct for the embedded ARM build. Guarded with `#if __SIZEOF_POINTER__ == 4`.

### `Libs/Header/SDK/Messages/CommandMessages.hpp`
18 struct-size `static_assert`s that assume 32-bit pointer layout. Same rationale as above — guarded with `#if __SIZEOF_POINTER__ == 4`.

### `Libs/Source/Wrappers/StdLibWrappers.c`
The Linux branch used `std::strncpy` in a `.c` file. `std::` is a C++ namespace and GCC 13 rejects it in C translation units. Replaced with plain `strncpy`.

## Notes

- All changes are backward-compatible. Windows simulator and ARM embedded builds are unaffected.
- The `MessageBase` / `CommandMessages` size assertions remain active for ARM (`__SIZEOF_POINTER__ == 4`) where binary layout matters.
- On x86-64, the structs are larger but the simulator runs both GUI and service in the same process, so both sides see the same layout.
- Full rationale documented in `Docs/SDK_LINUX_FIXES.md`.

## Test

Built and ran the TouchGFX simulator on Ubuntu 24.04 (GCC 13.3, x86-64) inside a Docker dev container. GUI renders correctly, screen navigation works.

### DIFF
--- a/Docs/SDK_LINUX_FIXES.md
+++ b/Docs/SDK_LINUX_FIXES.md
@@ -1,0 +1,94 @@
+# UNA SDK — Linux Simulator Compatibility Fixes
+
+This document records the changes made to the SDK to enable the TouchGFX simulator
+to build on x86-64 Linux (Ubuntu 24.04, GCC 13). All changes are backward-compatible:
+the embedded ARM build is unaffected and Windows simulator builds continue to work.
+
+---
+
+## Summary of changes
+
+| File | Issue | Fix |
+|------|-------|-----|
+| `Libs/Header/SDK/Simulator/OS/OS.hpp` | `#include <array>` only in `#ifdef _WIN32` block; `OS::Queue<T,N>` uses `std::array` | Add `#include <array>` to `#else` (Linux) branch |
+| `Libs/Header/SDK/Simulator/Kernel/Mock/App.hpp` | `#include <windows.h>` unconditional; nothing in the file actually uses it on Linux | Guard with `#ifdef _WIN32` |
+| `Libs/Header/SDK/Simulator/Kernel/Mock/Logger.hpp` | `GetTickCount64()` is Win32-only | Add portable `GetTickCount64()` wrapper using `clock_gettime(CLOCK_MONOTONIC)` for `#ifndef _WIN32` |
+| `Libs/Source/Simulator/Kernel/Mock/System.cpp` | `GetTickCount64()` and `Sleep()` are Win32-only | Add local wrappers using `clock_gettime` and `usleep` for `#ifndef _WIN32` |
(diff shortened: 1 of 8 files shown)


AvatarSD:
Hi @RegOrton ,

Thank you for your effort in running the simulator on Linux. Really impressive!

We're currently focusing on getting the basic simulator functionality working on Windows (as most of our hardware teammates are using it), but we also greatly value and support the Linux platform.
Running the simulator on Linux is very valuable, and we welcome any contributions to your effort.

Regards,
Denys and the UNA Watch team

tobymurray:
These changes generally look pretty innocuous - it would be amazing to get Linux support with fewer hoops. Is there a path to getting some or all of these merged? Are there less objectionable pieces of this PR we can get merged if we cut it up? Or is Una not interested at this point in time at all on the Linux front?

It feels like if Linux isn't supported we're going to end up with a Linux-supporting fork which is great that that's possible but a shame from a community building aspect.

rryles:
Rebased onto current `main` and pushed to `UNAWatch/una-sdk:pr-66-linux-simulator-rebased`. Replacement PR: https://github.com/UNAWatch/una-sdk/pull/123

rryles:
I've rebased these changes and incorporated them in #123 so closing this PR.

PR #70, opened 2026-03-31
Simulator: add sensor layer mock handlers in KernelMessageDispatcher

RegOrton:
## Summary

- `REQUEST_SENSOR_LAYER_GET_DEFAULT`, `REQUEST_SENSOR_LAYER_CONNECT`, and `REQUEST_SENSOR_LAYER_DISCONNECT` were falling through to the `default:` case in `KernelMessageDispatcher::appMsgHandler()`, so they were silently ignored
- This caused `SensorConnection::subscribe()` to never receive a `SUCCESS` result, making all sensor connections fail in the Linux x86-64 simulator
- Add minimal mock responses (return SUCCESS, assign handle = type ID for GET_DEFAULT) so apps can connect sensors and inject mock data directly in simulator builds

## Test plan

- [ ] Build Linux simulator for an app that uses sensors (e.g. Running, Rowing)
- [ ] Confirm sensor `subscribe()` returns true and connect succeeds in logs
- [ ] Confirm app logic that depends on sensors runs (e.g. GPS mock data flows through)

Relates to the existing Linux compatibility work in PR #66.

### DIFF
--- a/Libs/Source/Simulator/App/KernelMessageDispatcher.cpp
+++ b/Libs/Source/Simulator/App/KernelMessageDispatcher.cpp
@@ -7,6 +7,7 @@
 
 #include "SDK/Simulator/App/KernelMessageDispatcher.hpp"
 #include "SDK/Messages/CommandMessages.hpp"
+#include "SDK/Messages/SensorLayerMessages.hpp"
 
 #include <cstring>
 #include <memory>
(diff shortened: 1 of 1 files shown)


rryles:
Hi,

Since this PR was opened, main has picked up a fuller simulator sensor path for the same problem:

KernelMessageDispatcher now handles the sensor-layer message types explicitly (including REQUEST_SENSOR_LAYER_GET_DEFAULT, CONNECT, DISCONNECT, plus list/descriptor) using ::Sensor::Manager and the existing mSrvSensorListener / mMessageMgr wiring, instead of leaving them unhandled.
Instance::SensorLayer::getInstance().init() (already called from the dispatcher’s run()) registers the simulator sensor drivers with Sensor::Manager, so the dispatcher can resolve defaults/handles and complete connects/disconnects in a way that matches the rest of the simulator architecture.
When we rebased this branch onto current main, the patch became fully redundant with what is already in the tree (resolving the rebase conflict by keeping main’s KernelMessageDispatcher.cpp is equivalent to keeping your intent - real handling instead of silent drops - integrated via Sensor::Manager rather than the minimal “always SUCCESS” mock).

So we’re not rejecting the idea; we’re folding it in via the newer upstream implementation. We really appreciate you taking the time to trace the failure mode and propose a fix.

I'm closing this as superseded by main; if anything still fails on Linux please let us know and we can work on resolving it.

Thanks again for your contribution.

PR #102, opened 2026-05-01
fix(simulator): remove double destructor in MessageManager::releaseMessage

tobymurray:
Without this fix, pressing the exit key (key 4 / R2 / BACK) crashes the
simulator on shutdown:

```
  pure virtual method called
  terminate called without an active exception
  Aborted (core dumped)
```

The cause: releaseMessage() called msg->~MessageBase() explicitly before delete msg, invoking the virtual destructor twice. After the first call the vptr is reset to MessageBase's vtable; the subsequent delete dispatches through the base-class vtable and hits pure-virtual stubs in the derived message type's destructor.

Remove the explicit call so delete invokes the virtual dtor exactly once. operator delete is a no-op pool allocator and is unaffected.


rryles:
Hi @tobymurray 

Good spot and thanks for the fix. 

We really appreciate the time you've taken in helping to improve the SDK.

Regards,
Ross and the UNA Watch team

PR #104, opened 2026-05-01
docs: align project structure trees with actual layout

tobymurray:
Both the README and sdk-overview tree diagrams listed top-level Port/, Simulator/, and Visual Studio/ directories that don't exist. Port and Simulator actually live under Libs/{Header/SDK,Source}/, and the ThirdParty list omitted touchgfx and tinycbor. Update both trees to match the on-disk layout.


rryles:
Thanks for the corrections to our docs @tobymurray

PR #106, opened 2026-05-01
chore: use HTTPS for the Examples submodule URL

tobymurray:
Switch .gitmodules from the SSH form (`git@github.com:UNAWatch/una-apps.git`) to HTTPS (`https://github.com/UNAWatch/una-apps.git`) so that anonymous and unauthenticated checkouts can fetch the submodule without an SSH key.

Potentially too-descriptive motivation
----------
The current SSH URL silently breaks every entry path that does not have a GitHub SSH key configured:

  - First-time evaluators following the README
  - CI runners (the existing `tutorials-ci.yml` does not init the submodule today, but anything that does will hit this)
  - Docker image builds
  - Anonymous mirroring / archiving tools

This is the same direction the project already chose for the SDK itself: PR #56 documents HTTPS as the recommended clone option for the top-level repo, so aligning the submodule with that default is just making the two consistent. README.md still lists eight example apps (Alarm, Cycling, Hiking, ...) that all live in this submodule, so any new contributor who clones the SDK and tries those will hit the SSH failure today.

Pushing over HTTPS works fine in modern git (credential helpers cache auth, gh auth login handles 2FA/passkeys), so this change does not require maintainers to abandon SSH if they prefer it.

For maintainers who push to una-apps and want to keep using SSH 
----------
Two equivalent options, neither of which requires touching .gitmodules:

1. Per-repo override (this checkout only, this submodule only):
```
     git -C Examples remote set-url origin git@github.com:UNAWatch/una-apps.git
```
   Or before the submodule has been initialised:
```
     git config submodule.Examples.url git@github.com:UNAWatch/una-apps.git
     git submodule sync Examples
     git submodule update --init Examples
```

2. Global rewrite (all GitHub HTTPS URLs become SSH on this machine, for every repo, every submodule):
```
     git config --global url."git@github.com:".insteadOf "https://github.com/"
```


rryles:
Hi Toby,

You raise a very good point about the problems caused by using SSH for the submodule, and your proposed fix is valid.

I've decided to go further, and completely remove the submodule. Instead it is now a plain directory under the main una-sdk repository. Hopefully this addresses the issues you raised, and makes it easier for first time contributors to get started.

I'll therefore be closing this PR, but appreciate you submitting it none the less.

Best wishes,
Ross

tobymurray:
Yep - that resolves the same tension so no qualms about closing this, happy to see it.

PR #125, opened 2026-05-13
build(simulator): add -no-pie to Linux linker options for tutorial apps

tobymurray:
Adds the -no-pie flag to the Linux linker_options in all six tutorial una/Makefiles (HelloWorld, Sensors, Buttons, Files, Images, ScrollMenu). The change is inside the ifeq ($(UNAME), Linux) branch and does not affect the Windows build.

Without this change, building the simulator on modern Linux toolchains fails at link time. The pre-built lib/linux/libtouchgfx.a shipped with the SDK is not compiled with -fPIE, while contemporary GCC defaults to PIE for executables (Ubuntu 18.04+, Debian 10+, Fedora 23+). Linking non-PIE archive objects into a PIE executable produces relocation errors such as:

  relocation R_X86_64_32 against \`.rodata' can not be used when making a PIE object; recompile with -fPIE

With -no-pie, the executable is linked as a position-dependent binary, which matches the build mode of the pre-built libtouchgfx.a and lets the link succeed.


rryles (approved):
Looks like a pragmatic fix to get these working on Linux. Thanks.

PR #126, opened 2026-05-13
chore(tutorials): bump generated image banners to TouchGFX 4.26.1

tobymurray:
When the vendored TouchGFX SDK was bumped from 4.26.0 to 4.26.1 in 218a5b3 (chore(touchgfx): bump to 4.26.1), only files under `ThirdParty/touchgfx/` were updated. The per-tutorial `generated/` trees still carried the version banner stamped by the 4.26.0 toolchain. A fresh build of any tutorial after the SDK bump dirties the working tree with these regenerated headers.

The 4.26.1 imageconvert produces byte-identical output to 4.26.0 for every image in teh tutorials. This PR captures only that label sync across all 6 tutorials so anyone running `make assets` against the in-tree 4.26.1 toolchain stops getting dirty-tree noise from image regeneration.

The previous revision of this PR also included regenerated `generated/fonts/` output. @rryles Windows test on that revision flagged that `fontconvert` reverses some of the byte changes when re-run on Windows.

- `imageconvert` - output appears to be platform independent
- `fontconvert` -output appears to be platform dependent. There are two separate native binaries (`Thirdparty/touchgfx/framework/tools/fontconvert/build/{linux,win}/fontconvert.out` and produce slightly different rasterizations.
  - speculating this is due to the bundled FreeType builds

This revision keeps only the `imageconvert` subset, which is platform independent.

Continues to not modify any source inputs.


rryles:
Hi @tobymurray 

I ran a quick test on Windows and unfortunately saw exactly the type of churn you feared. When I regenerated the Buttons tutorial on top of this branch, I got a diff that appears to be roughly the reverse of yours. For example the changes to `Font_Poppins_Light_60_2bpp_0.cpp` were reverted. So it seems as if this is more of an OS difference than a TouchGFX version difference.

I know some people have had success running TouchGFX Designer under bottles/wine. Ultimately that would be the most likely path to end-to-end Linux development. I don't know if you've tried that approach?

Thanks,
Ross

tobymurray:
I feared as much! I'll rework this to only do the `4.26.0` -> `4.26.1` bump (as that should be platform agnostic and something that should be done every time the tool version bumps, I would assume).

> Ultimately that would be the most likely path to end-to-end Linux development. I don't know if you've tried that approach?

I haven't tried that approach, no - the simulator seems to work just fine on linux with a handful of small patches. I've been using it recently with no issues I've noticed so far. I have a branch [here](https://github.com/tobymurray/una-sdk/tree/linux-simulator). I've been trying to upstream the less contentious bits to narrow the diff as much as possible.

Obviously the change in the generated assets is something a Linux native build has to face, and there are a few others I have on my fork but my general take is that this kind of all works on Linux without too much fuss? I haven't been motivated yet to use bottles/wine.

rryles (approved):

PR #123, opened 2026-05-13
Linux x86-64 simulator build compatibility (rebase of #66)

rryles:
## Summary

This PR is a **replacement for [#66](https://github.com/UNAWatch/una-sdk/pull/66)** with the same intent (Linux x86-64 TouchGFX simulator / GCC 13 fixes), **rebased onto current `main`** and conflicts resolved upstream (e.g. `Kernel/Mock/App.hpp` removed on `main`; `RequestSystemSettings` size assert aligned with 64-byte layout on 32-bit).

Original contribution: [@RegOrton](https://github.com/RegOrton) — thank you.

Closes nothing by default; consider closing [#66](https://github.com/UNAWatch/una-sdk/pull/66) if this merges.

## Notes

- See `Docs/SDK_LINUX_FIXES.md` on this branch for rationale.
- Branch: `pr-66-linux-simulator-rebased` on `UNAWatch/una-sdk`.

PR #127, opened 2026-05-13
fix(simulator): make Windows.h includes Linux-safe

tobymurray:
Six tutorial Model.cpp files (HelloWorld, Sensors, Buttons, Files, Images, ScrollMenu) and the SDK's SwTimer.cpp unconditionally included a Windows header inside a section guarded only by SIMULATOR. The header doesn't exist on Linux, so the simulator build fails there with "fatal error: Windows.h: No such file or directory" before any source has a chance to compile.

Follows the same approach as #123 (rebase of #66), which fixed the equivalent issue in the SDK-side mock layer. That PR's commit (e2b8fc6) wrapped the `Kernel/Mock/App.hpp` `windows.h` include in `#ifdef _WIN32`; this one applies the same pattern to the remaining unguarded Windows headers in the tutorials and SwTimer.cpp.

In the tutorial Model.cpp files the include sits next to <chrono> and <ctime> in a block already wrapped in `#if defined(SIMULATOR)`. The simulator code in these files calls no Windows API, so the header is dead even on Windows, but the minimum-risk fix is to wrap it in `#ifdef _WIN32` / `#endif` rather than delete it.

SwTimer.cpp's `#include <windows.h>` is dead unconditionally — the file's only function uses `std::chrono::steady_clock` and references no Win32 symbol. It is removed outright rather than guarded.

Impact on Windows: none. Windows.h is still included where it was before; <windows.h> in SwTimer.cpp was not contributing any symbol used by the file.

Impact on Linux: the simulator can compile these translation units.

### DIFF
--- a/Docs/Tutorials/Buttons/Software/Apps/TouchGFX-GUI/gui/src/model/Model.cpp
+++ b/Docs/Tutorials/Buttons/Software/Apps/TouchGFX-GUI/gui/src/model/Model.cpp
@@ -11,7 +11,9 @@
 
 #if defined(SIMULATOR)
     #include "touchgfx/canvas_widget_renderer/CanvasWidgetRenderer.hpp"
+    #ifdef _WIN32
     #include "Windows.h"
+    #endif
     #include <chrono>
     #include <ctime>
 #endif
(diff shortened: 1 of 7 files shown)


rryles (approved):
Looks good. Thanks.

PR #128, opened 2026-05-14
fix(simulator): Directory::readNext returns one entry per call

tobymurray:
Aligns Mock::Directory::readNext with the IDirectory contract.

readNext is a cursor: one entry per call, advancing each time, returning false when there's nothing left. reset is a pure rewind that returns without reading and leaves item alone. That's now spelled out in the IFileSystem.hpp doxygen (#163).

The mock doesn't do that. It loops over the whole directory in one do/while and always returns true:
```
do {
    safe_strcpy(item.name, findData.cFileName, ...);
    ...
} while (FindNextFileA(handle, &findData));
return true;
```

So the first call after open() runs through every entry and leaves item on the last one. Later calls re-run the body with stale findData and keep handing back that same last entry, with no way to signal "done", so while (dir->readNext(info)) never ends.

The fix keeps the Win32 calls (FindFirstFileA, FindNextFileA, FindClose) and adds a hasEntry flag for whether findData holds an entry that hasn't been returned yet:

- open() sets hasEntry from the open result, so the first readNext already has FindFirstFileA's entry ready.
- readNext() with reset reopens to rewind and returns right away without reading; the next call returns the first entry. Otherwise it returns false when there's no entry, copies findData into item, advances with FindNextFileA, and updates hasEntry.
- close() clears hasEntry so a stale entry can't survive open/close/open.

Same entries (including "." and "..") in the same order as before.

Nothing in the SDK, the Examples submodule, or the tutorials calls readNext today, so this is safe to land.

### DIFF
--- a/Libs/Header/SDK/Simulator/Kernel/Mock/FileSystem.hpp
+++ b/Libs/Header/SDK/Simulator/Kernel/Mock/FileSystem.hpp
@@ -97,6 +97,7 @@ private:
     HANDLE handle;
     WIN32_FIND_DATAA findData;
     bool isOpenFlag = false;
+    bool hasEntry = false; // 'true' when findData holds an entry not yet returned by readNext
     char pathBuffer[SDK::Interface::IFileSystem::skMaxPathLen]; // Buffer to hold the path without prefix
 
     const char *mPathPrefix;
--- a/Libs/Source/Simulator/Kernel/Mock/FileSystem.cpp
+++ b/Libs/Source/Simulator/Kernel/Mock/FileSystem.cpp
@@ -258,6 +258,7 @@ bool Directory::open()
     std::string searchPath = path + "/*";
     handle = FindFirstFileA(searchPath.c_str(), &findData);
     isOpenFlag = (handle != INVALID_HANDLE_VALUE);
+    hasEntry = isOpenFlag;
     return isOpenFlag;
 }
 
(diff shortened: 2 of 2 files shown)


rryles:
Hi Toby. Sorry we've been a bit unresponsive lately - we've been busy behind the scenes adding new features.

Heads-up: I've rebased this onto latest main, so it should be mergeable as-is.

One bit of context that helps here: we just merged a docs change (#163) that fleshes out the `IDirectory::readNext` doxygen, and it's now in your base. It pins down the two things the old one-liner left ambiguous — that `readNext` is a one-entry-per-call cursor (which your fix correctly implements), and crucially the `reset` semantics, which you flagged you weren't sure about. So the contract is now fully specified in the header itself.

The newly-documented `reset` behavior is:

> When `reset` is `true`, the cursor is first rewound to the start of the directory. A reset call does NOT read an entry: `item` is left unmodified and the return value only reflects whether the rewind succeeded. The next call with `reset` `false` returns the first entry.

Against that contract there's one thing left to tweak in this PR. The current `reset` branch reopens the directory but then falls through into the copy block:

```cpp
if (reset) {
    FindClose(handle);
    open();              // loads first entry into findData, hasEntry = true
}

if (!hasEntry) return false;
// ... copies the first entry into item, then advances past it:
hasEntry = (FindNextFileA(handle, &findData) != 0);
return true;
```

So a reset call *consumes* the first entry instead of just rewinding. A caller doing the common "rewind, then iterate" pattern…

```cpp
dir->readNext(info, /*reset=*/true);   // expected: pure rewind
while (dir->readNext(info)) { ... }    // expected: starts at entry #0
```

…would silently skip the first entry. Since `open()` already loads the first entry into `findData` and sets `hasEntry = true`, the reset branch just needs to return right after reopening:

```cpp
if (reset) {
    FindClose(handle);
    return open();   // rewind only; the next readNext returns the first entry
}
```

That matches the documented contract — reset rewinds and returns without reading, leaving `item` untouched, and the next non-reset call surfaces the first entry.

Everything else looks great — the `hasEntry` flag, clearing it in `close()`, and the per-call cursor logic are all spot on. With that one adjustment to the reset branch this is good to go. And thanks again for the thorough write-up on the original PR; the ambiguity was on our side, and #163 should make the contract clear for the next person too.

tobymurray:
> Sorry we've been a bit unresponsive lately

Busy times! More responsive than Garmin, so call that a win.

I've rebased and updated the PR, I think this matches what you're expecting now? No qualms with edits being made if it doesn't though.

rryles (approved):
Looks perfect now. Thanks.

PR #134, opened 2026-05-18
TouchGFX Bug reproduction for drawPartialBitmap Y-clipping bug

tobymurray:
**Do not merge**

This is a quirky one that is ultimately a bug with TouchGFX. It's definitely a bug on Linux, this branch is intended to be a disposable, easy-to-reproduce case to see if it also impacts Windows. I'll open a bug with STM regardless. I ran into this trying to get maps loading on the simulator.

If someone on Windows is able to check out this branch and run this, update with the outcome, then this is the expectation:

| Expected | Bug |
|--------|--------|
| pr-screenshot-fix | pr-screenshot-bug | 

When you draw an ABGR2222 dynamic-external bitmap with LCD::drawPartialBitmap and part of the bitmap extends past the dirty rect on the Y axis, the renderer writes those rows starting at the dirty rect's top instead of at the requested screenY, producing a duplicate/misplaced copy at the wrong vertical position. Hits any code that scrolls bitmaps in/out of view, anchors them near screen edges, or clips them vertically for any other reason.

  Triggers (all must hold):
  - Bitmap format Bitmap::ABGR2222 (or sibling 8bpp 2222 - static analysis shows identical compiled code in all four)
  - Bitmap registered via Bitmap::dynamicBitmapCreateExternal (other registration types untested)
  - Routed through LCD::drawPartialBitmap, or LCD::blitCopy with a full-bitmap blitRect
  - Source rect extends past the dirty rect on the Y axis (screenY + height > dirty.y + dirty.height, or screenY < dirty.y)
  - screenY ≠ dirtyRect.y (typically screenY ≠ 0)

  Non-triggers: X axis is handled correctly — screenX can be negative or off-screen without effect. Non-zero screenY alone is harmless when the bitmap fits inside the dirty rect (verified at screenY = 1).

### DIFF
--- a/Docs/Tutorials/HelloWorld/Software/Apps/TouchGFX-GUI/gui/include/gui/main_screen/MainView.hpp
+++ b/Docs/Tutorials/HelloWorld/Software/Apps/TouchGFX-GUI/gui/include/gui/main_screen/MainView.hpp
@@ -3,6 +3,47 @@
 
 #include <gui_generated/main_screen/MainViewBase.hpp>
 #include <gui/main_screen/MainPresenter.hpp>
+#include <touchgfx/widgets/Widget.hpp>
+#include <touchgfx/Bitmap.hpp>
+#include <touchgfx/hal/HAL.hpp>
+
+// Minimal reproduction for a Y-clipping defect in
+// LCD::drawPartialBitmap, observed on the Linux simulator with the
+// prebuilt libtouchgfx.a 4.26.1. The widget draws a 50x50 ABGR2222
+// dynamic external bitmap at screen (60, 220) so the source rect
+// extends 30 px past the bottom of the 240x240 dirty rect, exercising
+// the renderer's clipping path.
+//
+// Expected: a single red strip about 20 rows tall near the bottom of
+// the widget; the upper part shows the default HelloWorld content.
+//
+// On the defective build, a second red strip also appears near the
+// top of the widget because the clipping path writes destination rows
+// starting at dirtyRect.y instead of screenY. Two red strips means
+// the bug reproduces on your platform; a single strip near the bottom
+// means it does not.
+class DrawPartialBitmapBugRepro : public touchgfx::Widget
+{
+public:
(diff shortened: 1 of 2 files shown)


rryles:
Confirmed on **Windows** — and on **device** too.

**Windows simulator (MSVC, `LCD8bpp_ABGR2222` @ 240×240):** built this branch's HelloWorld tutorial sim and the bug reproduces — two red strips, one at the expected bottom position (screenY=220) and a spurious second one at the top of the canvas. The top strip is the bottom-overhang rows being written at `dirtyRect.y` (0) instead of `screenY` (220), exactly as you described.

**On device (STM32U595, same `LCD8bpp_ABGR2222` panel):** cross-compiled the same repro to a `.uapp`, installed it, and the watch shows the identical two-strip artifact.

So this is **not Linux-specific** — it reproduces across all three targets (Linux, Windows, hardware). That lines up with your static-analysis note: the sim links the x86 `libtouchgfx.a` and the firmware links the ARM `libtouchgfx-float-abi-hard.a`, both built from the same TouchGFX 4.26.1 source, and the defect is in the shared `LCD8bpp_ABGR2222` clipping path — so platform is irrelevant. Definitely worth raising with STM as a framework bug; happy to add the Windows/device details to that report if useful.

Triggers all held as you listed: ABGR2222 dynamic-external bitmap, via `drawPartialBitmap`, source rect overhanging the dirty rect on Y, with `screenY ≠ dirtyRect.y`. Thanks for the self-contained repro — made this trivial to verify.

tobymurray:
Awesome - thanks for taking the time (and reproducing on the device as well, looking forward to doing so myself). It doesn't look like there's necessarily a bug tracker for STM, just a community forum but I'll try a post there.

tobymurray:
"filed a bug" (created a post in the community forum) here: https://community.st.com/stm32-mcus-touchgfx-and-gui-33/rawpartialbitmap-draws-y-overhang-rows-at-dirty-rect-top-instead-of-screeny-abgr2222-external-bitmap-touchgfx-4-26-1-166187

PR #156, opened 2026-06-09
fix(activity): derive lap/session avg speed from distance/time

rryles:
## Problem

The lap (and session) average pace stored in the FIT file is wrong: for a 1 km / 1 mi auto-lap the reported average pace does not match the lap duration.

`ActivityWriter` serializes whatever `speedAvg` the service hands it, and its own comments document the FIT definition: `avg_speed = total_distance / total_timer_time`. But the services fed in `VariableCounter::getAverage()` / `getLapAverage()` — the **arithmetic mean of the instantaneous speed samples**. That sample mean drifts away from distance/time because:

1. **Sub-threshold filtering biases it fast.** Samples below `minValid` (0.5 m/s) are excluded from the average, but those seconds still count toward the lap/session duration. Every slow stretch (stop light, turnaround) makes the average read faster than `distance / time`.
2. **Speed and distance are independent.** On GPS apps, distance is GPS displacement while speed is Doppler instantaneous speed — their sample mean need not reconcile with displacement/time.

So `distance` and `duration` were exact, but `avg_speed` was not consistent with them. The on-watch displayed average pace, derived from the same counter, was wrong the same way.

## Fix

Derive lap, session and live average speed as `distance / active-timer-time` (the data the services already track) via a small `speedFromTotals()` helper, so distance, duration and pace always agree.

Applied to **Running, Cycling, Treadmill, and Hiking**. Treadmill's *session* total was already derived this way in `finalizeActivity()`; its per-lap and live values now match it (`max_speed` / per-record data are left as recorded, unchanged).

## Additional fixes folded in

- **Hiking: stop recording every lap twice.** `Hiking::Service::saveLap()` pushed each lap onto `mSummary.laps` twice (a brace-initialised entry *and* an explicit `LapSummary`), so the summary listed double the laps. Removed the redundant push and fed the survivor from the `lapTime`/`lapDistance` locals. (Reported by CodeRabbit on this PR.)
- **Simulator link fix (Cycling & Hiking).** Their `Application.vcxproj` never compiled `ImuRunningCadence.cpp` (constructed by `InstanceSensorLayer`), so the host simulators failed to link; added the missing `ClCompile` (+ the matching `.filters` grouping). Running/Treadmill already had it.

> Note: the broader Running/Treadmill `.vcxproj.filters` drift cleanup is intentionally **not** here — it's a separate cosmetic PR (#157).

## Verification

**Builds:** all four host simulators build and link clean (Debug/x86, VS 2022 MSBuild).

**Behaviour (driven through the actual GUI simulators with scripted button input; values read from the production-written `Activity/summary.json`):**

*Running* — `time:147 s, distance:768.045 m, speed_avg:5.2248 m/s, pace_avg→3:11/km, lap_count:2`
- session: 768.045 / 147 = 5.2247 m/s = `speed_avg`; `pace_avg` = 1/speed = 3:11/km (matched the on-screen AVG PACE)
- lap 1: `pace 0.19384 s/m × dist 624.226 m = 121.0 s = dur` ✓
- lap 2: `pace 0.177581 s/m × dist 140.781 m = 25.0 s = dur` ✓

Each lap's average pace now **exactly equals `lap_duration / lap_distance`** — the behaviour the report asked for.

*Hiking* — `time:57 s, distance:190.762 m, speed_avg:3.34671 m/s, lap_count:2`
- session: 190.762 / 57 = 3.347 m/s = `speed_avg` ✓
- `lap_count:2` with two **distinct** laps `[{31s,99.8m},{25s,85.3m}]` — pre-fix this run would have written **4** duplicated entries, confirming the dedup fix.

## Scope / safety

SDK-app-side only; no kernel change. Net behavioural effect: average pace now equals `lap_time / lap_distance` for distance auto-laps, matching user expectation and the FIT spec, and Hiking no longer double-counts laps.

### DIFF
--- a/Examples/Apps/Cycling/Software/Apps/TouchGFX-GUI/simulator/msvs/Application.vcxproj
+++ b/Examples/Apps/Cycling/Software/Apps/TouchGFX-GUI/simulator/msvs/Application.vcxproj
@@ -55,6 +55,7 @@
     <ClCompile Include="$(SdkPath)\Libs\Source\Simulator\Components\Sensors\SensorPressure.cpp"/>
     <ClCompile Include="$(SdkPath)\Libs\Source\Simulator\Components\Sensors\Imu\ImuWristMotion.cpp"/>
     <ClCompile Include="$(SdkPath)\Libs\Source\Simulator\Components\Sensors\Imu\ImuStepCounter.cpp"/>
+    <ClCompile Include="$(SdkPath)\Libs\Source\Simulator\Components\Sensors\Imu\ImuRunningCadence.cpp"/>
     <ClCompile Include="$(SdkPath)\Libs\Source\Simulator\Components\ComponentSimulator.cpp"/>
     <ClCompile Include="$(SdkPath)\Libs\Source\Simulator\Components\Simulator\GpsStepCounterSimulator.cpp"/>
     <ClCompile Include="$(SdkPath)\Libs\Source\Simulator\Components\SensorListener.cpp"/>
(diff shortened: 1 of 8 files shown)


rryles:
Thanks @coderabbitai — good catch. Fixed in e0aa10e: removed the duplicate `LapSummary` push in `Hiking::Service::saveLap()` and fed the surviving brace-initialised entry from the `lapTime`/`lapDistance` locals, matching the other three apps. Each lap is now recorded once. Verified the Hiking host simulator builds and links clean.

sdvsaienko (approved):

PR #164, opened 2026-06-13
fix(simulator): set isHidden/isSystem in FileSystem::objectInfo

tobymurray:
`objectInfo` never set `item.isHidden` or `item.isSystem`, so callers read uninitialized values. `readNext` already sets both from the `WIN32_FIND_DATA` attributes; `objectInfo` now does the same.

I missed this in #128, noticed now that I was updating everything with all the work that has been done on the repo.

## Summary by CodeRabbit

* **Tests**
  * Enhanced mock FileSystem implementation to properly simulate hidden and system file attributes, enabling more realistic filesystem behavior during testing operations.


rryles (approved):
Nice one. Thanks.

PR #166, opened 2026-06-15
feat(sensor): add SpO2 to the sensor type enum

tobymurray:
The hardware has an optical blood-oxygen sensor, but there's no `SDK::Sensor::Type` value for it, so it can't be referenced through the sensor layer. This is a proposal for how that might look — happy to have you take it in whatever direction fits the SDK. What I'm really after is *some* representation of SpO2 so it can eventually be wired up in the simulator.

What's here:

- **`SPO2 = 0x000000F1`** in `SensorTypes.hpp`. I slotted it one past `PPG` (`0xF0`) in the optical block, since SpO2 comes off the PPG path — mirroring how `HEART_RATE` (`0x41`) sits next to `HEART_BEAT` (`0x40`). The exact value is yours to decide.
- **`SensorDataParserSpo2.hpp`** — SpO2 reads as a processed scalar (saturation % + trust level) rather than a raw waveform, so I modeled the parser on `HeartRate`: `getSaturation()` / `getTrustLevel()` plus the usual validity and timestamp accessors.
- **`SensorsLayer.md`** — table row and a short usage section to match.

Header-only; nothing changes for existing types.

## Summary by CodeRabbit

* **New Features**
  * Added support for an Optical **SPO2** sensor, providing **saturation (%)** and **trust level** measurements.

* **Documentation**
  * Updated the Sensors Layer documentation with the new **SPO2 (0xF1)** sensor type, including field descriptions and example accessor usage.

### DIFF
--- a/Docs/SensorsLayer.md
+++ b/Docs/SensorsLayer.md
@@ -35,6 +35,7 @@ All available sensor types are defined in [`SDK::Sensor::Type`](../Libs/Header/S
 | Motion | GESTURE_RECOGNITION | 0xD0 | Discrete gesture events | No | - |
 | Daily | ACTIVITY | 0xE0 | Active minutes (minutes) | Yes | DURATION (u32 ms) - 1 |
 | Optical | PPG | 0xF0 | Photoplethysmogram | No | - |
+| Optical | SPO2 | 0xF1 | Blood-oxygen saturation (SpO2) | Yes | SATURATION (float %), TRUST_LEVEL (float) - 2 |
 | Optical | ECG | 0x100 | Electrocardiogram | No | - |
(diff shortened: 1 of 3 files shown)


rryles (approved):
Approving — clean, correctly-scoped API stub, and the `saturation % + trust` shape matches what the sensor layer will publish, so it's good to build against.

Two notes for when it lands:

1. **`0xF1` is locked as the committed ABI value for `SPO2`.** We're treating it as fixed from here, so the simulator and the eventual kernel publisher can depend on it without churn.

2. **SpO2 will be background-sampled, not a live/continuous stream.** Readings arrive sparsely (measured automatically when conditions allow) rather than as a steady feed like `HEART_RATE`. Consumers — the simulator included — should treat each reading as a point-in-time sample and lean on the timestamp (i.e. surface how old the value is) rather than assume it's current. The `getTimestamp()` / `getTimestampUs()` accessors already cover that.

Thanks for the contribution.

PR #167, opened 2026-06-15
Add BeatProbe: runnable diagnostic for the HEART_BEAT vs PPG question

tobymurray:
This is theoretically merge-able, but probably not useful to do so beyond capturing the diagnostic output.

I'm particularly interested in the ability for the watch to measure HRV, as that underpins a whole bunch of the "training readiness" metrics. You may already know the answers to this, but if not is someone with a watch able to run and report? Without a device, I'm unable to tell if the firmware emits `HEART_BEAT` (0x40) as timed peak events (RR intervals are gaps between beat timestamps, so HRV is cheap streaming stats), or does the raw PPG (0xF0) waveform come out (so RR needs an on-watch peak detector)?

If you happen to know the answers to these questions, amazing, otherwise please try running the tutorial on a device and report the outcome:

### Questions (may have answers or may need hardware)
1. on a worn, HR-calibrated watch, does subscribing to `Type::HEART_BEAT` (0x40) deliver ~one event per heartbeat?
     - I would expect the answer would be, "yes", "events but not per beat" or "no events"
2. If yes to 1, is each beat's timestamp (`mTimeStamp` ms + `mTimeStampUs` us) the beat-detection instant or the IPC delivery time?
3. Does the beat event carry a payload (RR/IBI value or quality flag), or is it a bare event?
4. If `HEART_BEAT` is _not_ emitted, is a continuous `PPG` (0xF0) waveform available, and at what rate / how many channels?

### Real watch experiment
Run this branch (check out the branch, build the Sensors app, flash, wear the device for ~90 seconds, then grep `BEAT-EVENT` in the debug UART. Theoretically it prints a self-interpreting verdict (obviously I don't have a watch to try it on).

If possible, sharing the full log would be appreciated, otherwise just the relevant output would answer the question this PR poses

## Release Notes

* **New Features**
  * Added a diagnostic probe that gathers heart-beat and PPG evidence for ~90s, computes beat timing and quality metrics, periodically summarizes progress, and prints a final A/B/C verdict.

* **Improvements**
  * Integrated the probe into the Sensors service so it automatically runs during normal operation and forces a final report when the app stops.

* **Documentation**
  * Updated the Sensors tutorial with build/deploy instructions and a guide for reading the diagnostic log and verdict meanings.

### DIFF
--- a/Docs/Tutorials/Sensors/Software/Libs/Header/BeatProbe.hpp
+++ b/Docs/Tutorials/Sensors/Software/Libs/Header/BeatProbe.hpp
@@ -1,0 +1,268 @@
+/**
+ ******************************************************************************
+ * @file    BeatProbe.hpp
+ * @brief   Drop-in diagnostic that answers the "beat-event question":
+ *          does HEART_BEAT (0x40) deliver timed peak events (=> HRV is cheap),
+ *          or must RR be recovered from the raw PPG (0xF0) waveform
+ *          (=> HRV needs an on-watch peak detector)?
+ *
+ * It observes four streams that the Sensors tutorial already subscribes to:
+ *   - HEART_BEAT     (0x40)  beat peak events  -> RR = delta of timestamps
+ *   - PPG            (0xF0)  raw optical waveform
+ *   - HEART_RATE     (0x41)  BPM + trust       -> cross-check vs RR-derived HR
+ *   - TOUCH_DETECT   (0x140) worn / unworn      -> gates the "no data" case
+ *
+ * No GUI, no filesystem, no FIT. Output is log-only and self-interpreting. It logs a
+ * raw line per beat (timestamp + inter-beat interval) and ~1 PPG line/s as it goes, plus
+ * a summary block every 15 s, one at 90 s, and one on exit, each ending in VERDICT A/B/C.
+ * Even a truncated or early-terminated log is conclusive. Grep for "PROBE" or "BEAT-EVENT".
+ *
+ * Integer math only (no <cmath>, no float printf) so it is robust on the
+ * MCU's newlib. Single-instance: declare one file-static BeatProbe.
+ ******************************************************************************
+ */
+#pragma once
+
+#include <cstddef>   // SensorDataBatch.hpp uses size_t without including this itself
+#include <cstdint>
+
+#include "SDK/SensorLayer/SensorDataBatch.hpp"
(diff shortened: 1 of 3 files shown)


rryles:
Hi Toby,
HRV is definitely something we want to do. However, it's going to take some work.

> on a worn, HR-calibrated watch, does subscribing to Type::HEART_BEAT (0x40) deliver ~one event per heartbeat?

At the moment I'm afraid it's "no events". The heart rate detection isn't detecting individual beats - it's more of a frequency domain algorithm.

> If HEART_BEAT is not emitted, is a continuous PPG (0xF0) waveform available, and at what rate / how many channels?

Getting a raw PPG waveform could be a route to HRV. For measuring HR (which is all the PPG sensor does today), the data is sampled at 20 Hz and is a single channel. This is at the low end for HRV extraction. We probably want to switch to a higher frequency (and higher power consumption) mode temporarily when measuring HRV.

We are also looking into the possibility of getting HRV to be calculated by the PPG chip itself.

So bottom line, not today, but we are working on it.

I should also be clear - this will only ever be HRV at rest. Measuring HRV while moving (e.g. mid exercise) can't be done optically - it has to be an electrical measurement. Overnight or resting morning HRV is perfect for training readiness/recovery metrics though.

tobymurray:
@rryles - helpful response, thank you. Keen to see if on-chip HRV pans out in a useful way.

One avenue that sidesteps both the issues you mentioned that I currently regularly use is an external chest strap (electrical, so HRV during exercise is viable). `AccessoryMessage` and `Kind::HRM` support that via `HEART_RATE_EX`. The beat-to-beat data is inaccessible to an app though. I believe the RR intervals ride in the same 0x2A37 notification the watch already reads for the BPM, but by the time it hits the the SDK it's been collapsed.

Could the kernel surface the RR array as a new frame?

Zooming out, the strap is one instance of a broader decision - Bluetooth is noticeably less accessible than the rest of the platform. Since `RequestPepare` only takes a bitmask of kernel-defined kinds, there doesn't appear to be a way to point the watch at any arbitrary BLE device.

Would it be viable to expose a generic, "connect to this device, subscribe to this characteristic, hand me the raw bytes" pass through in the SDK? Buys a ton of community functionality and provides a very natural progression from SDK-implemented experimental functionality being "promoted" into the firmware once proven mature/valuable.

Happy to contribute on that front, extremely invested in anything that makes the watch feel more like "my" hardware.

PR #177, opened 2026-07-03
feat(simulator): POSIX implementation of the mock FileSystem

tobymurray:
Open to hear if this is contentious - the CI is all Linux but obviously the test coverage is Windows-focused because of the TouchGFX tooling. This is, I think, the last notable piece that is upstream-able to support Linux for running the simulator. The builds are not identical, and the Linux will have to stay constrained to a fork, but if we're able to upstream this it's very close to it Just Works on.

Mock::FileSystem was Windows-only (Win32 + FILETIME), so the simulator did not build on Linux. Add a POSIX path in the #else branch without touching the Windows functionality.

No Windows regression: all new logic is confined to the #else path, so the Win32 branches compile as before. The shared changes are limited to an RAII destructor, deleted copy ops, and zero-initializing the read/write counts; on Windows each is either inert (std::fstream already self-closes) or a strict improvement (it closes the previously-leaked Directory HANDLE and prevents a double-close), never a regression.

The POSIX File is built on a raw fd rather than std::fstream: IFile has a single read/write position that matches an fd offset (fstream's separate get/put positions do not), which also avoids the fstream append-mode bug where seek()+write() appended instead of overwriting. It handles EINTR/short I/O, ftruncate, fsync, and closes on destruction. Directory uses opendir/readdir plus stat: isDir via S_ISDIR, all ObjectInfo fields initialized, "." and ".." skipped, rewinddir on reset.

Add Tests/Host coverage, POSIX-guarded so it runs on the existing Linux host-tests CI job: mkdir/write/read/size, seek+overwrite, truncate, objectInfo (poisoned struct and trailing slash), directory enumeration, rename/copy/remove, and fd/DIR* release under a lowered RLIMIT_NOFILE. First automated coverage for this file.

## Summary by CodeRabbit

* **New Features**
  * Added/strengthened POSIX behavior for the simulator’s mock filesystem, including nested directory creation and directory enumeration.
  * Improved metadata reporting for files and directories (size, timestamps, and proper trailing-slash handling).
* **Bug Fixes**
  * More reliable file I/O, including correct success checks, robust read/write retry logic, and accurate seek/truncate behavior.
  * Improved resource management by ensuring proper handle cleanup and safe reopen behavior.
* **Tests**
  * Added extensive host-side coverage for core filesystem operations, metadata correctness, enumeration edge cases, and handle-leak/reopen scenarios.

### DIFF
--- a/Libs/Header/SDK/Simulator/Kernel/Mock/FileSystem.hpp
+++ b/Libs/Header/SDK/Simulator/Kernel/Mock/FileSystem.hpp
@@ -5,7 +5,11 @@
 #include <fstream>
 #include <string>
 #include <sstream>
+#ifdef _WIN32
 #include <windows.h>
+#else
+#include <dirent.h>
+#endif
 
 #include "SDK/Interfaces/IFileSystem.hpp"
 
@@ -20,7 +24,11 @@ public:
 
     File(const char *prefix, const char *relativePath);
 
-    ~File() = default;
+    ~File() override;
+
+    // Owns a raw handle (fd on POSIX); non-copyable to avoid a double close.
+    File(const File &) = delete;
+    File &operator=(const File &) = delete;
 
 
     virtual void setPath(const char *path) override;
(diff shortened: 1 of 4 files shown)


rryles:
Hi Toby. This is a great bit of work and not contentious at all. We'd love to better support linux as a development environment. As you say, it's TouchGFX that is the obstacle to this. I'm hopeful we can get to a place where Linux based development doesn't feel second class, and this is a step in that direction.

PR #182, opened 2026-07-08
fix(apps): pin distance auto-laps to the km/mi grid (no drift)

rryles:
## Problem

Distance auto-lap (e.g. "auto lap every 1 km") drifts. The lap fired when the
**per-lap** distance counter crossed the target, then `resetLap()` re-based the
counter at the current sensor reading — discarding the overshoot past the
boundary. At 1 Hz sampling each tick advances several metres, so every lap ran
`target + overshoot`, producing three symptoms from one root cause:

1. **Boundaries drift** off the km/mi grid as the overshoot accumulates.
2. Laps record as e.g. **1.01 km** instead of a clean km.
3. Per-lap **average pace disagrees with the lap duration** (a 1.006 km lap
   reports `time / 1.006 km`).

## Fix

**`fix(apps): pin distance auto-laps to the km/mi grid`**
- Adds `MonotonicCounter::advanceLap(span)`: closes the current lap as exactly
  `span` and carries the overshoot into the next lap by shifting the lap base
  forward (both active and includes-pauses accumulators) instead of zeroing it.
- Running, Hiking, Cycling and Treadmill now record each distance auto-lap as
  exactly the target and carry the remainder, so boundaries stay pinned to the
  grid. Because pace is derived as `target / lapTime`, the reported lap pace
  now equals the lap duration by construction. `saveLap()` gained a
  `float autoLapDistanceM = 0.0f` parameter; time, manual and interval laps
  keep using `resetLap()` and are unchanged.
- Session total distance is untouched; recorded laps still sum to N×target plus
  the final partial lap.

**`fix(metrics): don't drop the first movement when a counter starts at zero`**
- A pre-existing quirk surfaced while testing: `add()` decided whether the base
  was seated using the value-sentinel `mBaseValue == T{} && mValueActive == T{}`.
  When the first sample equals `T{}` (distance starting at 0) the sentinel
  stayed true, so the next call re-seated the base and the first real movement
  was dropped. Now keyed off `mHasData`. Non-zero-start behaviour is unchanged;
  the `advanceLap` carry is unaffected.

**`fix(metrics): make resetLap gap-free; test mixed auto/manual laps`**
- `resetLap()` zeroed the lap base and re-seated on the *next* sample, dropping
  the sub-sample sliver between a lap event and the following sample — so manual
  (and time / interval) laps lost a few metres each from lap accounting (session
  total unaffected). It now closes the lap in place by advancing each lap base by
  what the lap consumed (active and includes-pauses independently, pause-safe),
  matching the gap-free behaviour `advanceLap` already gave auto-laps. The
  now-unreachable lap re-seat branch in `add()` is removed.
- Adds a **mixed auto+manual regression test**: verifies a manual lap re-syncs
  the auto grid to the manual point (next auto one full target later, not on the
  original grid) and that recorded laps + final partial sum to the total (no gap).

## Tests

New host tests in `SDK/Tests/Host/metrics/MonotonicCounter_test.cpp` (8 tests):
overshoot carry, active/total carry across a pause, span clamp, gap-free reset,
zero-start regression, a 2000-sample multi-lap drift regression, a mixed
auto+manual end-to-end test, and a reset-during-pause consistency test.
**Full SDK host suite: 142/142 pass.** All four app service ELFs build for ARM.

## Note / follow-up

On an unfiltered GPS spike where a single 1 Hz sample jumps ≥2× the lap target,
the code emits a short burst of degenerate ~1 s laps (one per subsequent sample)
as the grid catches up, rather than one oversized lap. No loop/hang — at most one
lap per tick. GPS distance is filtered upstream so this is largely theoretical;
happy to add a multi-boundary skip if we want to harden it.

## Summary by CodeRabbit

* **New Features**
  * Added improved distance-based auto-lap handling so auto-laps can align precisely to the configured lap target.
* **Bug Fixes**
  * Corrected lap distance reporting for distance auto-laps, including better management of boundary overshoot to prevent later laps from being skewed.
  * Improved long-run stability to reduce accumulated lap distance drift when repeatedly splitting by distance.
* **Tests**
  * Expanded coverage for distance auto-lap/overshoot behavior, including clamping and mixed manual/auto lap sequences.

### DIFF
--- a/Examples/Apps/Cycling/Software/Libs/Header/Service.hpp
+++ b/Examples/Apps/Cycling/Software/Libs/Header/Service.hpp
@@ -142,7 +142,7 @@ private:
     void sendInitialInfoToGui();
     void startTrack(std::time_t utc);
     void processTrack();
-    void saveLap();
+    void saveLap(float autoLapDistanceM = 0.0f);
     void stopTrack(bool discard);
     void pauseTrack(bool pause);
     void buildPartialSummary();
--- a/Examples/Apps/Cycling/Software/Libs/Sources/Service.cpp
+++ b/Examples/Apps/Cycling/Software/Libs/Sources/Service.cpp
@@ -709,11 +709,17 @@ void Service::processTrack()
         mSessionNotEmpty = true;    // Session has at least one record
         mLapNotEmpty = true;        // Lap has at least one record
 
-        bool switchLap = false;
+        bool  switchLap        = false;
+        float autoLapDistanceM = 0.0f;  // >0 marks a grid-aligned distance auto-lap
         switch (mLapDivSource) {
-        case LapDivSource::DISTANCE:
-            switchLap = mDistanceCounter.getLapValueActive() >= Settings::Alerts::Distance::toMeters(mSettings.alertDistanceId, mIsImperial);
+        case LapDivSource::DISTANCE: {
+            const float target = Settings::Alerts::Distance::toMeters(mSettings.alertDistanceId, mIsImperial);
+            if (mDistanceCounter.getLapValueActive() >= target) {
+                switchLap        = true;
+                autoLapDistanceM = target;
+            }
             break;
+        }
         case LapDivSource::TIME:
             switchLap = static_cast<uint32_t>(mTimeCounter.getLapValueActive()) >= Settings::Alerts::Time::toSeconds(mSettings.alertTimeId);
             break;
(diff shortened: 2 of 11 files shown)


rryles:
@coderabbitai fullreview

rryles:
Thanks @coderabbitai. Addressed in a322ecb:

- **`advanceLap` span guard** — added a clamp so `span` can never exceed the current lap value (no negative accumulators). Callers already gate on `getLapValueActive() >= span`, so this is purely defensive hardening of the public API.
- **`AdvanceLapCarriesActiveAndTotal` didn't diverge active vs total** — the test now inserts a pause so `total > active` (704 vs 1004 → 4 vs 304 after `advanceLap`), actually exercising the independent carry. Added `AdvanceLapClampsOversizedSpan` for the new guard.

Deferring the third nit (**extract the duplicated grid-lap logic across the four Services**): these four activity apps are intentionally standalone copies that ship independently, and factoring out shared lap logic is a cross-app architecture change beyond the scope of this drift fix — better as its own refactor. Noted for a follow-up.

Full SDK host suite: 142/142; all four service ELFs build for ARM.

PR #194, opened 2026-07-15
feat(settings): implement time-of-day format setting

tobymurray:
Americans are starting to get watches (so jealous!) and the very first reactions I've seen are unfortunately incredulity around lack of support for different time formats.

This is intended to be backward compatible so I think should Just Work. Setting the actual configuration for 24h compact will need to be implemented in UNA's closed source firmware, but that gap has no negative impact on current users. The `RequestSystemSettings.timeFormat` field existed but was dead plumbing: the kernel dispatcher never populated it, no app read it, and every clock rendered hardcoded 24-hour time.

Replace the bool (which could only express two formats and had ambiguous polarity) with an explicit uint8_t-backed enum, SDK::Message::TimeFormat: Hour24 (0, default, '17:42'), Hour12 ('5:42'), Hour12Extended ('1742'). Same size as the bool, so RequestSystemSettings stays 64 bytes. When we're changing this, making it extensible is cheap - I am not aware of anyone voicing a desire to see 24 hour compact time but I leveraged it as an avenue to show how more formats could be supported (are there more than these 3?).

- Simulator: add `timeFormat` to `WatchSettings` and copy it into `REQUEST_SYSTEM_SETTINGS` responses in `KernelMessageDispatcher`.
- Apps (Cycling, Hiking, Running, Treadmill, Workout): carry the format through `SettingsUpd` -> `Service` -> `Mode`l -> `TrackView`, and render the status-face clock accordingly in `TrackFaceStatus::setTime`.

12-hour mode shows no AM/PM suffix: the status-face wildcard buffer and generated fonts have no A/P/M glyphs, and adding them means regenerating TouchGFX assets for every app. I am not set up to do so presently, so all formats use digits already present.

The Alarm app is untouched, the time wheel remains 24-hour. It is not wired up for settings at all, doing so feels like a different effort than this. On top of that, a 12-hour picker needs an AM/PM wheel and TouchGFX Designer layout changes.

### Default (24 hour, ISO-8601 extended)
time-format-24h-single-digit
time-format-24h

### 12 hour
time-format-12h

### 24 hour, ISO-8601 Compact
time-format-military

## Summary by CodeRabbit

* **New Features**
  * Added support for configurable 12-hour, 24-hour, and compact time displays.
  * Time-format settings now flow from device settings to the Cycling, Hiking, Running, Treadmill, and Workout interfaces.
  * Track screens display times using the selected format.

* **Bug Fixes**
  * Corrected time displays that previously always used a fixed 24-hour format.

* **Tests**
  * Added coverage for supported formats, edge cases, and fallback behavior.

### DIFF
--- a/Docs/Examples/Cycling-Architecture.md
+++ b/Docs/Examples/Cycling-Architecture.md
@@ -680,6 +680,7 @@ constexpr SDK::MessageType::Type MANUAL_LAP     = 0x0000000F;
 struct SettingsUpd : public SDK::MessageBase {
     Settings settings;
     bool     unitsImperial;
+    SDK::Message::TimeFormat timeFormat;
     uint8_t  hrThresholds[kHrThresholdsCount];  // C-array of 6
     uint8_t  hrThresholdsCount;
 };
@@ -703,7 +704,7 @@ public:
     Sender(const SDK::Kernel& kernel);
 
     // Service --> GUI
-    bool settingsUpd(Settings settings, bool units,
+    bool settingsUpd(Settings settings, bool units, SDK::Message::TimeFormat timeFormat,
                      const uint8_t (&thresholds)[kHrThresholdsCount],
                      uint8_t thresholdCount);
     bool time(std::tm localTime);
--- a/Docs/Examples/Hiking-Architecture.md
+++ b/Docs/Examples/Hiking-Architecture.md
@@ -682,6 +682,7 @@ constexpr SDK::MessageType::Type MANUAL_LAP     = 0x0000000F;
 struct SettingsUpd : public SDK::MessageBase {
     Settings settings;
     bool     unitsImperial;
+    SDK::Message::TimeFormat timeFormat;
     uint8_t  hrThresholds[kHrThresholdsCount];  // C-array of 6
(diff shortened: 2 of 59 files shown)


tobymurray:
As I look at my current Garmin watch I am just now noticing that it 0-pads the 24 hour time, that may be be more conventional.

rryles:
Hi Toby.
I don't have access to the shipping lists, but hopefully your watch is well on it's way to you.

Support for 12 hour time has been something we've been working on internally, and your work here helped inform the path we've taken. We did consider supporting more than two formats, but decided that we would stick to just 12 hour and 24 hour. Leading zeros and colons are relevant, but we decided they are a design choice for a watch face. We do plan to allow users to develop their own watch faces and it would be a bit of a burden if they had to support three or more different time formats on every face.

tobymurray:
That makes sense, as I was implementing it it felt a bit code-smell-y to conflate the 12 hour and 24 hour configuration with the specific representation. I do think an SDK utility to avoid every app needing to re-implement things may still be something worth distilling, I'll give that a shot and see what it looks like, may help lower the watch face burden.

PR #211, opened 2026-07-20
fix(ci): stop release notes with backticks breaking the Slack step

rryles:
`apps-v1.3.0-rc1` release CI failed at **Prepare Slack payload** (exit 127): a PR title in the auto-generated notes contained backticks (around `ELEV.`), and the step spliced the release body into the shell via `${{ steps.get_slack_params.outputs.body }}`, so the shell ran command substitution on the backticks.

The app builds, the release, and the asset upload all succeeded — only the Slack notification failed.

**Fix:** pass the release fields (body/name/tag/author/created/url) via an `env:` block and reference them as quoted shell vars (`"$BODY_RAW"`, `$SLACK_*`) instead of `${{ }}`-interpolating them into the script. Env values are literal, so backticks / `$(...)` in notes can't execute — this also closes the template-injection hole. Applied to both `apps-ci.yml` and `sdk-release.yml` (identical pattern).

---

**Also folds in two related hardening fixes** (from a follow-up review — same "a release-controlled string breaks the release job" class):

1. **Tag name no longer spliced into the shell.** `gh release edit`/`gh release view ${{ github.ref_name }}` are replaced with a job-level `env: TAG:` referenced as `"$TAG"`. Git ref names may legally contain backticks/`$`/`;`, so a hostile tag name could otherwise inject (low risk — needs push access — but cheap to close). Note kernel `ci.yml` triggers on `tags: ['*']`.
2. **Randomized `$GITHUB_OUTPUT` heredoc delimiter** for the release body (`body<<EOF` → `body<<EOF_<random>`). A release note line that is exactly `EOF` would otherwise terminate the heredoc early and fail/spoof the step. The `payload<<EOF` heredoc is left as-is (jq -c output is single-line — provably safe).

## Summary by CodeRabbit

* **Chores**
  * Improved release automation consistency across application and SDK releases.
  * Enhanced release metadata handling for Slack notifications, including release tags, authors, timestamps, links, and descriptions.
  * Improved reliability when formatting and publishing release notifications.

### DIFF
--- a/.github/workflows/apps-ci.yml
+++ b/.github/workflows/apps-ci.yml
@@ -140,14 +140,21 @@ jobs:
 
       - id: prepare_slack_payload
         name: Prepare Slack payload
+        env:
+          BODY_RAW: ${{ steps.get_slack_params.outputs.body }}
+          SLACK_NAME: ${{ steps.get_slack_params.outputs.name }}
+          SLACK_TAG: ${{ github.ref_name }}
+          SLACK_AUTHOR: ${{ steps.get_slack_params.outputs.author }}
+          SLACK_CREATED: ${{ steps.get_slack_params.outputs.created_at }}
+          SLACK_URL: ${{ steps.get_slack_params.outputs.url }}
         run: |
-          BODY="${{ steps.get_slack_params.outputs.body }}"
+          BODY="$BODY_RAW"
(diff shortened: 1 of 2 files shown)


rryles:
@coderabbitai full review

PR #212, opened 2026-07-21
ci(simulator): add non-gating Linux simulator build matrix

tobymurray:
The time has come! As far as I can tell, the simulator and all examples build and run on Linux out of the box.

This adds an informational GitHub Actions workflow that builds the TouchGFX SDL2 desktop simulator for every example app and tutorial on x86-64 Linux and boots each headlessly. This is stuff not covered by either apps-ci.yml (ARM) nor host-tests.yml (GTest).

Self-discovers projects (simulator/gcc/Makefile) and scopes the matrix by change: app-only PRs build only the touched project; shared changes (Libs/\*\*, ThirdParty/\*\*) or push/dispatch build all; a PR whose base cannot be resolved falls back to building all. A broken build or boot turns the check red (so the regression is visible) but never blocks a merge, since it is not a required status check. Per-project logs and the binary are uploaded.

## Summary by CodeRabbit

* **Tests**
  * Added an automated Linux Simulator build and smoke-test workflow.
  * Runs selectively for pull requests and across all simulator projects for pushes or manual executions.
  * Verifies simulator startup in a headless environment and reports results in the workflow summary.

* **Chores**
  * Build logs, simulation logs, and simulator binaries are automatically retained as downloadable artifacts for 14 days.

### DIFF
--- a/.github/workflows/linux-simulator.yml
+++ b/.github/workflows/linux-simulator.yml
@@ -1,0 +1,182 @@
+name: Linux Simulator (non-gating)
+
+on:
+  pull_request:
+    paths:
+      - 'Libs/**'
+      - 'ThirdParty/**'
+      - 'Examples/Apps/**'
+      - 'Docs/Tutorials/**'
+      - '.github/workflows/linux-simulator.yml'
+  push:
+    branches: [ main, develop ]
+    paths:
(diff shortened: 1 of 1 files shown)


rryles:
Nice one!

PR #219, opened 2026-07-23
refactor(apps): retire the per-app Sender with SDK helper

tobymurray:
This is an exercise in bloat reduction, trying to make it so any new example has less to copy/paste.

Every app hand-rolled a `CustomMessage::Sender` whose methods were all the same allocate/set/send/release sequence. These were all copies of what it appears `make_msg` was introduced to accomplish.

`allocateMessage` and `make_msg` now forward constructor arguments, and a new `send_msg<T>(kernel, ...)` does the allocate/send/release in one call. Messages are plain structs again: each gets a constructor that fills its own fields (the memcpy/strncpy/clamp cases carried over unchanged) and no longer holds a kernel reference. Call sites become
`SDK::send_msg<CustomMessage::TrackStateUpd>(kernel, state)`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability and consistency of GUI↔service messaging across Alarm, Cycling, HR Monitor, Hiking, Running, Treadmill, and Workout apps.
  * Alarm playback/control, snoozing, and alarm list updates now propagate more consistently, including active-alarm startup behavior.
  * Track/lap lifecycle events, summaries, sensor/telemetry updates, calibration status, and interval-training notifications are forwarded more reliably.
  * Startup/resume GUI state (initial settings/summaries and live telemetry like time/battery/GPS) stays correctly synchronized.

### DIFF
--- a/Examples/Apps/Alarm/Software/Apps/TouchGFX-GUI/gui/include/gui/model/Model.hpp
+++ b/Examples/Apps/Alarm/Software/Apps/TouchGFX-GUI/gui/include/gui/model/Model.hpp
@@ -64,7 +64,6 @@ public:
 private:
     ModelListener*        modelListener;
     const SDK::Kernel&    mKernel;
-    CustomMessage::Sender mSrvSender;
 
     // IGuiLifeCycleCallback
     void onStart()   override;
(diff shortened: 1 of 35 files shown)


tobymurray:
> Handle failed SDK::send_msg results

This PR stays consistent with the legacy design which similarly did not handle failed results. Doing so is more of a design/behavior question about whether these are best effort or intended to be handled robustly and outside the scope of this structural change.

PR #222, opened 2026-07-25
fix(simulator): move HALSDL2 include from Mock/System.hpp to System.cpp

tobymurray:
`Mock/System.hpp` includes TouchGFX's `HALSDL2.hpp` but doesn't use anything from it. The only use is `SystemGUI::exit()` in `System.cpp`. The include forces every simulator component that pulls in the `Mock/System.hpp` (e.g., sensor manager, sensor sources, kernel dispatcher, etc) to depend on TouchGFX and the SDL2, so none of them can build headless.

Move the include to the TU that needs it and give the two files that were silently relying on the transitive dependency their own: `<cstlib>` for `AppMemory.hpp` and `<cmath>` + `std::fabs` for `SensorBatteryLevel.cpp`. As noted in the review, `<cstdlib>` is deliberately _not_ the fix for the `SensorBatteryLevel.cpp` as it binds `::abs(int)`, truncating the float delta to 0 (making a mess of the publish interval as well)

### DIFF
--- a/Libs/Source/Simulator/Components/Sensors/SensorBatteryLevel.cpp
+++ b/Libs/Source/Simulator/Components/Sensors/SensorBatteryLevel.cpp
@@ -16,6 +16,7 @@
 
 #include "SDK/Simulator/Components/Sensors/SensorBatteryLevel.hpp"
 #include "SDK/SensorLayer/DataParsers/SensorDataParserBatteryLevel.hpp"
+#include <cmath>
 #include <SDK/Simulator/Kernel/Mock/System.hpp>
 #include "SDK/Simulator/OS/OS.hpp"
 #include "SDK/Simulator/Components/ComponentSimulator.hpp"
@@ -104,7 +105,7 @@ void BatteryLevel::sensorRefresh()
     float level = mBattLevel.getBattLevel();
 
     OS::MutexCS cs(mDataMutex);
-    if (mPrevLevel > 0 && abs(mPrevLevel - level) < 0.1) {
+    if (mPrevLevel > 0 && std::fabs(mPrevLevel - level) < 0.1) {
         LOG_DEBUG("EXIT\n");
         return;
     }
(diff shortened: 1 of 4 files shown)


Libs/Source/Simulator/Components/Sensors/SensorBatteryLevel.cpp
```diff
@@ -20,6 +20,7 @@
 #include "SDK/Simulator/OS/OS.hpp"
 #include "SDK/Simulator/Components/ComponentSimulator.hpp"
 
+#include <cstdlib>
```
rryles: Nit, but I'd take it before merge: `<cstdlib>` is the wrong header for the call this unblocks.

The only consumer is the float comparison at line 107:

```cpp
if (mPrevLevel > 0 && abs(mPrevLevel - level) < 0.1) {
```

`mPrevLevel` is `float` (`SensorBatteryLevel.hpp:49`) and so is `level`, so this wants `<cmath>` — `<cstdlib>`'s job is the integer overloads. It compiles today only because MSVC and libstdc++ both hoist `<cmath>`'s floating-point overloads into the global namespace from `<cstdlib>`; that isn't guaranteed, and libc++ is less generous about `::abs(float)`. Where it doesn't hold, unqualified `abs` binds `abs(int)`, the difference truncates to integer granularity, and sub-1.0 changes get silently swallowed by the early return — wrong behaviour rather than a build error, which is the annoying failure mode.

Suggest:

```cpp
#include <cmath>
```

and

```cpp
if (mPrevLevel > 0 && std::fabs(mPrevLevel - level) < 0.1f) {
```

For what it's worth I checked the current state locally (MSVC 14.51, `/std:c++17`): without the new include this is a hard `error C3861: 'abs': identifier not found`, so the include you added is load-bearing rather than cosmetic, and with it `abs(float - float)` does bind the 4-byte float overload. So there's no behaviour change as the PR stands — this is purely about not leaning on that resolution, which seems worth doing in a PR whose whole point is removing toolchain-dependent include luck.

PR #224, opened 2026-07-28
fix(tutorials): correct 90° display rotation on device

tobymurray:
I installed HelloWorld on my physical watch (yay, finally got it!) and it renders rotated 90°

image

It looks like all six TouchGFX tutorials (HelloWorld, Buttons, Images, Files, Sensors, ScrollMenu) ship `application.config` with `"layout_rotation": "90"`, which the generator bakes into `FrontendApplicationBase` as `setDisplayOrientation(ORIENTATION_PORTRAIT)`.

The bug is invisible in the simulator as it renders to its own window.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected the default display orientation across TouchGFX tutorial applications.
  * Updated image layout settings to 0° rotation, preventing unintended 90° rotation.

* **Documentation**
  * Updated the Buttons, Files, HelloWorld, Images, ScrollMenu, and Sensors tutorial projects to use landscape orientation.


rryles:
### Heads-up: the CI `.uapp` artifacts from this PR need a kernel built from `main`

If you test the artifacts from this PR's CI on a **1.2.0** watch, both apps will appear to launch and
then leave the watch on an `App PID 0x........` screen with no button responding — recoverable only by
a ~16 s hard reset. That is **not** a problem with this branch, and not related to the rotation fix.

The serial log gives it away immediately:

```
HelloWorld.SRV  -E- system::una_init_kernel:99: Kernel not supported. Minimum 3, got 2
Kernel          -E- App.Kernel::exit:100 : [HelloWorld.SRV] Abnormal exit. Status -4
HelloWorld.GUI  -E- system::una_init_kernel:99: Kernel not supported. Minimum 3, got 2
```

**Why:** GitHub Actions `pull_request` builds the **merge commit**, not the branch tip. This PR's CI
built `refs/pull/224/merge` = `44af141` (the hash embedded in the artifact filename,
`HelloWorld_1.3.0-26-44af141.uapp`), and that merge includes `main`'s
`feat(abi): bump KERNEL_INTERFACE_VERSION 2 -> 3`. So:

| | `KERNEL_INTERFACE_VERSION` |
|---|---|
| this branch's tip | 2 |
| `44af141` (what CI built and published) | 3 |
| kernel 1.2.0 on-device | provides 2 |

`system.cpp`'s gate is `gIKernel->version < KERNEL_INTERFACE_VERSION`, so 2 < 3 fails and both the
service and GUI processes take the `exitA(-4)` path. This affects **any** app built from a PR against
post-bump `main` — nothing tutorial-specific.

**To test on hardware:**
- build the `.uapp` locally with the ABI set to 2 so it still runs on 1.2.0 or 1.3.0 kernel.

**Two follow-ups, both outside this PR's scope:**

1. The unexitable-screen behaviour is a kernel bug worth fixing on its own:
   `UserAppView::handleKeyEvent()` has its entire body commented out, and the view's `mPid` is never
   cleared when a process dies — so the `handleTickEvent()` escape (`mPid == 0`) never fires even
   though the log confirms both processes terminated. Any app that dies after launch therefore wedges
   the watch.
2. An ABI-incompatible app should say "this app needs a newer watch firmware" rather than silently wedging —
   the kernel already logs `Abnormal exit. Status -4`, so it has what it needs. This is the gap
   `Docs/Versioning-and-Releases.md` §3.3 flags (compatibility should be checkable before install;
   app-start is too late). Once ABI 3 ships, any 1.3.0 user sideloading a newer app hits it.

The rotation fix itself is now verified on-device: layout upright and the button bitmaps correct.

PR #228, opened 2026-07-29
fix(sensor): recover a lost sensor subscribe (GPS position missing when an activity starts before fix)

rryles:
## Problem

A field user reported activities uploading with **no map**. A sample FIT confirmed it: a full 38-minute run — HR, speed, cadence, **5.3 km of distance with three 1-mile auto-laps** — but **zero GPS position records** for the entire activity. Since distance on 1.2.0 is the haversine of *valid* GPS coordinates, the GNSS clearly had real fixes throughout; only the app-recorded position was missing. It reproduces (rarely) when the user starts the activity **before GPS fix is acquired**.

## Root cause

Per-record position is written only when the app's own `GPS_LOCATION` subscription yields a fix. The app subscribes to `GPS_LOCATION` once, early, at GUI start (so acquisition begins on the pre-activity screen). `SensorConnection::subscribe()`/`connect()` use a 100 ms message timeout; during the busy app-startup window the `RequestDefault` round-trip can exceed it, so `subscribe()` fails, `mHandle` stays `0`, `matchesDriver()` never matches, and the app ignores location for the **whole session** — with no retry. Distance/speed still record via the kernel's own in-kernel `GPS_LOCATION` listener (the `GPS_DISTANCE` sub-sensor), which masks the outage — so the activity looks complete but has no map.

Also, `connect()` used `send(100) || ok()`; `sendMessage()` returns `true` even on a response *timeout*, so it latched "connected" on a timed-out ack and never recovered.

## Fix

- **`SensorConnection::connect()`**: `mIsConnected = send(100) && ok()` — connected only on a real SUCCESS reply, so a timeout/FAIL leaves it retryable.
- **`SensorConnection::disconnect()`**: send `RequestDisconnect` whenever a handle is held (not only when `mIsConnected`) — a connect whose ack timed out can still have registered the listener kernel-side; gating on `mIsConnected` leaked it (and the duty-cycled hardware behind it) until app stop.
- **Apps (Running/Cycling/Hiking/Workout)**: retry the subscription until it takes.
  - `GPS_LOCATION`: retried from the 1 Hz loop, gated on a new `mGpsWanted` intent (set at `onStartGUI`, cleared in `disconnect()`), so acquisition still starts on the pre-activity screen but the GNSS is never re-woken on the post-activity summary screen.
  - Track sensors: `connectSensors()` made idempotent (per-sensor `isConnected()` guard) and pumped from `processTrack()` (runs only while the track is active). Workout uses no GPS, so it gets only this part.
  - A `WARNING` is logged when the startup subscribe is lost and an `INFO` when the retry recovers it, so field/monitoring logs show when the race actually fires.

## Testing

- New host tests `Tests/Host/sensorlayer/SensorConnection_test.cpp` (7 cases) covering subscribe timeout/FAIL recovery, the `send && ok` connect flag, the disconnect hardening, and reconnect-without-resubscribe. The race can't be reproduced on demand on hardware, so these drive the client logic deterministically with a scripted comm that mirrors the device (`sendMessage` returns `true` on a response timeout).
- **Verified with a negative control**: reverting the two `SensorConnection` changes turns exactly the three fix-dependent tests red.
- Full host suite: **175/175**. All four activity service ELFs link for ARM.

## Notes / out of scope

- Treadmill has the same one-shot `connectSensors()` but is being folded into Running via activity variants, so it is intentionally left untouched.
- On-device confirmation is optional now that the host tests are the deterministic acceptance; the new recovery log lets the field confirm the fix engaging.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved GPS reliability by automatically retrying connections when startup acknowledgements are missed.
  - Added sensor connection recovery during active tracking without reconnecting sensors that are already online.
  - Prevented stale sensor connections after interrupted or timed-out connection attempts.
  - Ensured GPS does not reconnect after the activity or app has stopped.

- **Tests**
  - Added comprehensive coverage for sensor connection failures, retries, timeouts, and cleanup behavior.

### DIFF
--- a/Examples/Apps/Cycling/Software/Libs/Header/Service.hpp
+++ b/Examples/Apps/Cycling/Software/Libs/Header/Service.hpp
@@ -114,6 +114,7 @@ private:
     Track::State mTrackState          = Track::State::INACTIVE;
     bool         mPreviousGpsFixState = false;
     bool         mGpsInitialConnectFailed = false;  ///< GPS_LOCATION subscribe lost the startup ack race; the retry logs the recovery.
+    bool         mGpsWanted = false;                 ///< GPS_LOCATION should stay connected (pre-activity + active track); cleared in disconnect() so the retry never re-wakes the GNSS post-activity.
     bool         mSessionNotEmpty     = false;
     bool         mLapNotEmpty         = false;
     Track::Data  mTrackData{};
--- a/Examples/Apps/Cycling/Software/Libs/Header/Service.hpp
+++ b/Examples/Apps/Cycling/Software/Libs/Header/Service.hpp
@@ -113,6 +113,7 @@ private:
     LapDivSource mLapDivSource        = LapDivSource::OFF;
     Track::State mTrackState          = Track::State::INACTIVE;
     bool         mPreviousGpsFixState = false;
+    bool         mGpsInitialConnectFailed = false;  ///< GPS_LOCATION subscribe lost the startup ack race; the retry logs the recovery.
     bool         mSessionNotEmpty     = false;
     bool         mLapNotEmpty         = false;
     Track::Data  mTrackData{};
--- a/Examples/Apps/Cycling/Software/Libs/Sources/Service.cpp
+++ b/Examples/Apps/Cycling/Software/Libs/Sources/Service.cpp
@@ -287,12 +287,13 @@ void Service::disconnect()
     }
 
     // The activity is over (stopTrack) or the app is stopping: GPS is no longer
-    // wanted, so the run() retry must not re-wake it. Release it if still up.
+    // wanted, so the run() retry must not re-wake it. Release unconditionally --
+    // disconnect() is a no-op if never subscribed, and firing it whenever a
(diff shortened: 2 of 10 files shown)


rryles:
Thanks @coderabbitai — both addressed in 1ada811:

- **`mIsConnected` not reset on `make_msg` failure** (`SensorConnection::disconnect`): moved `mIsConnected = false` out of the `if (request)` block so it clears regardless of allocation.
- **GPS disconnect gated on `isConnected()` in the apps** (major): dropped the `if (mSensorGpsLocation.isConnected())` guard in `Service::disconnect()` for Running/Cycling/Hiking — good catch, it was defeating the `disconnect()` hardening in exactly the connect-ack-timeout case (`mIsConnected` false) it was added for. `disconnect()` is a safe no-op when never subscribed.

Validated: host suite 175/175 (incl. the 7 `SensorConnection` cases), all four service ELFs link for ARM.

sdvsaienko (approved):

PR #231, opened 2026-07-30
feat(sdk): app-pushed home-screen widget IPC channel

sdvsaienko:
## Summary
A new Service→kernel IPC channel that lets a running app surface a small live
element — a status label and/or a completion bar — on the kernel's home screen
for the duration of an ongoing activity (e.g. a Timer showing its countdown
while its own GUI is closed). This is **app-push**, distinct from glances (a
separate, kernel-driven app type).

This PR is the **SDK contract only**: the messages, an ergonomic helper, and
simulator acceptance. Kernel-side rendering/placement and the first consumer
(Timer) land separately.

## The contract
Three fire-and-forget requests (`0x0330`–`0x0332`):
- `REQUEST_WIDGET_START` — claim the single home-widget slot. Empty payload; the
  app is identified by its process.
- `REQUEST_WIDGET_UPDATE` — `{ uint32 shown; float percent; char text[64] }`.
  `shown` (a `WidgetShow` bitmask) authoritatively selects which fields to show
  now and carries render modifiers; `percent` is 0..100; `text` is UTF-8,
  NUL-terminated.
- `REQUEST_WIDGET_STOP` — release the slot.

Helper `SDK::HomeWidget` (`start` / `stop` / `update`) wraps the sends.

## Semantics & design
- **Replace, not merge:** each update fully specifies what's shown; fields whose
  `WidgetShow` bit is not set are hidden.
- **Extensible:** new capabilities are added by *appending* a field and/or a new
  `WidgetShow` bit — never reorder/remove/reuse. A bit either gates a field (e.g.
  an icon) or modifies how a shown field renders. Old and new senders and readers
  interoperate by ignoring bits they don't know.
- **Progress direction:** `WIDGET_PROGRESS_FROM_END` — a `WidgetShow` modifier bit
  for `percent` that fixes which point the fill is measured from. Default fills
  from the bar's *start* (a rising percent fills up, e.g. steps toward a goal);
  set pins the fill to the *end* (a falling percent empties from the start, e.g.
  a countdown sending its remaining share). Start/end are logical — the kernel
  maps them to the widget's layout and orientation.
- **Robust against bad input** (a POD over IPC can't validate the sender, so the
  reader is defensive):
  - `shown` is masked to known bits (helper *and* kernel) — unknown/garbage bits
    are dropped.
  - `percent` is clamped to `[0,100]` (NaN/inf → bounds).
  - every field is zero-initialised (placement-new ctor), so the worst a bad
    `shown` can do is blank a field — never a crash or over-read.
  - `text` is copied UTF-8-safe (never splits a multi-byte sequence).
- **Ownership:** a single slot, last-START-wins; only the current owner's
  UPDATE/STOP take effect; no owner stack (STOP clears the slot, it isn't handed
  back to a displaced app); the kernel clears the slot if the owner dies.

## Not in this PR (follow-ups)
- Kernel-side: home rendering, placement, per-app keying, clear-on-death, the
  display policy when multiple apps compete, and whether an active widget keeps
  the owning service scheduled.
- First consumer: the Timer service sending start/update/stop.

## Testing
- `static_assert(sizeof(RequestWidgetUpdate) == 104)` on the 32-bit ABI.
- Compiles under `arm-none-eabi-g++` (headers + helper, incl. a Cyrillic UTF-8
  label) and in the MSVC Win32 simulator (the Stopwatch app builds, exercising
  the new dispatcher cases).
- `copyUtf8` truncation/boundary behaviour covered by host tests (`Tests/Host`,
  10 cases: ASCII + multibyte truncation, mid-sequence back-off,
  all-continuation input, empty/null).

## Compatibility
New message ids only — ABI-safe for existing apps. No capability-flag or
kernel-interface-version change.

## Summary by CodeRabbit

* **New Features**
  * Added an SDK API to start, stop, and update a live home-screen widget (progress percent + UTF-8 label), including an option to anchor progress from the end.
* **Documentation**
  * Clarified widget update semantics, including how “shown” fields and progress anchoring work.
* **Bug Fixes**
  * Ensured UTF-8 label updates are bounded, truncation-safe, and always null-terminated.
* **Simulator**
  * Simulator now accepts widget start/update/stop requests and returns success.
* **Tests**
  * Added host unit tests covering UTF-8 bounded-copy and truncation edge cases.

### DIFF
--- a/Libs/Header/SDK/HomeWidget/HomeWidget.hpp
+++ b/Libs/Header/SDK/HomeWidget/HomeWidget.hpp
@@ -14,10 +14,12 @@ namespace SDK {
 namespace detail {
 
 /**
- * @brief Copy a UTF-8 string into a fixed buffer, NUL-terminated.
+ * @brief Copy a UTF-8 string into a fixed buffer, always NUL-terminated.
  *
- * Never splits a multi-byte sequence: if @p src does not fit, it is truncated
- * back to the last whole character so the buffer always holds valid UTF-8.
+ * @p src is assumed to be valid UTF-8. On truncation the copy backs off to a
+ * character boundary, so a valid @p src is never split mid-sequence. The routine
+ * does not validate @p src: malformed input is copied as-is (still bounded and
+ * terminated), never rejected.
  * @p dst holds at most @p cap - 1 payload bytes plus the terminator; a @p cap of
  * 0 writes nothing and a null @p src yields an empty string.
  *
--- a/Libs/Header/SDK/HomeWidget/HomeWidget.hpp
+++ b/Libs/Header/SDK/HomeWidget/HomeWidget.hpp
@@ -77,7 +77,7 @@ public:
     /** @brief Release the slot; the kernel removes the element. */
     bool stop()  { return send<SDK::Message::RequestWidgetStop>(); }
 
-    /** @brief Show a label and a 0..100 progress bar (both fields). */
+    /** @brief Show a label and a 0..100 progress bar filled from its start (both fields). */
     bool update(float percent, const char* text)
     {
         return update(SDK::Message::WIDGET_SHOW_TEXT | SDK::Message::WIDGET_SHOW_PERCENT,
@@ -89,6 +89,8 @@ public:
      *
      * Fields not in @p shown are hidden. @p text is UTF-8 and only used when
      * WIDGET_SHOW_TEXT is set; @p percent only when WIDGET_SHOW_PERCENT is set.
+     * Add WIDGET_PROGRESS_FROM_END to pin the fill to the bar's end -- a countdown
+     * that empties from the start.
      */
     bool update(uint32_t shown, float percent, const char* text)
     {
--- a/Libs/Header/SDK/HomeWidget/HomeWidget.hpp
+++ b/Libs/Header/SDK/HomeWidget/HomeWidget.hpp
@@ -1,0 +1,110 @@
+#ifndef HOMEWIDGET_HPP
+#define HOMEWIDGET_HPP
+
+#include <SDK/Kernel/Kernel.hpp>
+#include <SDK/Messages/CommandMessages.hpp>
+#include <SDK/Messages/MessageGuard.hpp>
+
+#include <cstddef>
+#include <cstring>
+
+namespace SDK {
+
+/**
+ * @brief Ergonomic wrapper for the home-screen widget a running app pushes.
+ *
+ * A running app (its Service) can surface a small live element on the kernel's
+ * home screen -- a status label and/or a completion bar -- for the duration of
+ * an ongoing activity. Bracket it with @ref start / @ref stop; refresh
+ * it with an @ref update whenever the app chooses (there is no fixed cadence). The
+ * app is identified by its process, so its name and icon come from the .uapp.
+ *
+ * An update is authoritative: the WidgetShow mask says exactly what the widget
+ * shows now; fields not in the mask are hidden. All sends are fire-and-forget.
+ * @code
+ *   SDK::HomeWidget widget(mKernel);
+ *   widget.start();
+ *   widget.update(42.0f, "02:35");                               // label + bar
+ *   widget.update(SDK::Message::WIDGET_SHOW_PERCENT, 42.0f, ""); // bar only
+ *   ...
+ *   widget.stop();
+ * @endcode
+ */
+class HomeWidget {
+public:
+    explicit HomeWidget(const SDK::Kernel& kernel) : mKernel(kernel) {}
+
+    /** @brief Claim the widget slot; the kernel starts showing the element. */
+    bool start() { return send<SDK::Message::RequestWidgetStart>(); }
+
+    /** @brief Release the slot; the kernel removes the element. */
+    bool stop()  { return send<SDK::Message::RequestWidgetStop>(); }
+
+    /** @brief Show a label and a 0..100 progress bar (both fields). */
+    bool update(float percent, const char* text)
+    {
+        return update(SDK::Message::WIDGET_SHOW_TEXT | SDK::Message::WIDGET_SHOW_PERCENT,
+                      percent, text);
+    }
+
+    /**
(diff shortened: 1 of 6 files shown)


rryles:
# Review — app-pushed home-screen widget IPC

**Verdict: solid, approvable.** The design is sound, well documented, and closely follows existing precedent (glance messages, the accessory ID block, the `static_assert`/`__SIZEOF_POINTER__` pattern). I found no correctness bugs. Below is one API footgun worth fixing before merge and a few minor points.

## What I verified (all check out)

- **Size assert**: `MessageBase` is 32 bytes under `pack(4)` on 32-bit; 32 + 4 + 4 + 64 = 104. ✓
- **Pool fit**: 104 bytes lands in the kernel's pool 2 (128-byte blocks, 50 count) — `RequestBuzzerPlay` at 116 already lives there, so no new pool pressure class. ✓
- **Message IDs**: `0x0330`–`0x0332` don't collide; they continue the de-facto `0x03xx` service-message block after accessories (`0x0320`–`0x0322`), and the gap `0x0323`–`0x032F` leaves accessories room. ✓
- **Fire-and-forget claim**: `MessageGuard::send()` defaults to `timeout = 0` → no completion wait. ✓
- **`copyUtf8`**: correct on the edge cases I traced — `cap == 1`, `nullptr` src, truncation landing exactly on a character boundary (no back-off), truncation mid-sequence (backs off to the lead byte and drops it), and pathological all-continuation-byte input (degrades to empty string; the `n > 0` guard prevents underflow). ✓
- **Percent clamp**: `(p >= 0) ? (p > 100 ? 100 : p) : 0` — NaN fails the first comparison → 0; +inf → 100. ✓
- **Simulator log**: `%.*s` with precision 64 bounds the read of a possibly non-NUL-terminated buffer — printf with a precision doesn't require termination. Correct defensive pattern. ✓

## Issues

### 1. Overload footgun in `HomeWidget::update` (worth fixing)

`WidgetShow` is an unscoped enum, so it converts implicitly to `float`. A caller who wants text-only and forgets the percent argument writes:

```cpp
widget.update(SDK::Message::WIDGET_SHOW_TEXT, "02:35");
```

This compiles silently — it resolves to the **two-arg** overload with `percent = 1.0f`, showing *both* fields with a 1% bar. Cheap compile-time fix:

```cpp
bool update(SDK::Message::WidgetShow, const char*) = delete;
bool update(uint32_t, const char*) = delete;   // catches OR-ed masks too
```

(The second one is needed because `WIDGET_SHOW_TEXT | WIDGET_SHOW_PERCENT` promotes to `uint32_t`.)

### 2. Simulator sanitize block is dead code

`KernelMessageDispatcher.cpp` sets `LOG_MODULE_LEVEL LOG_LEVEL_INFO`, and `LOG_DEBUG` compiles to `do {} while(0)` unless the module level is DEBUG. So in the `REQUEST_WIDGET_UPDATE` case, `shown`, `pct`, and `hasText` are computed and never used — likely `-Wunused-variable` warnings on the Linux build, and the "Sanitize as the reader contract requires" comment describes code that only feeds a compiled-out log. The existing glance-update case has the same issue (unused `gl`), so it's tolerated — but I'd either log at INFO, or drop the sanitize locals and keep the case a bare accept like START/STOP.

### 3. Minor / take-or-leave

- **Missing `<cstdint>`** in `HomeWidget.hpp` — `uint32_t` appears in the public signature but arrives only transitively via `CommandMessages.hpp`.
- **No test for `copyUtf8`** — it's the only nontrivial logic in the PR and exactly the shape of thing that grows regressions. The SDK repo has no host-test harness, so compile-acceptance is understandable, but the kernel repo's `Tests/Host` can include SDK headers; a small gtest there (boundary truncation, mid-sequence truncation, empty/null) would be a good follow-up.
- **Stale range doc**: the `MessageTypes.hpp` file header still documents type ranges (`0x1000` requests, `0x2000` responses…) that no real ID follows; this PR deepens the divergence. Pre-existing, not on this PR to fix, but worth a cleanup ticket.

## Things I'd normally flag but are fine here

- `REQUEST_*` naming for fire-and-forget messages matches the glance precedent (`REQUEST_GLANCE_UPDATE` is the same shape).
- The `#ifndef HOMEWIDGET_HPP` guard: SDK headers mix `#pragma once` and guards, so either is defensible.
- Zero-init constructor + reader-side masking contract is genuinely well thought through — the doc comments on `WidgetShow` explaining *why* the guarantee lives on the reader are the kind of thing that keeps this ABI-safe two years from now.
- Last-start-wins with no owner stack is the right call for a v1 single-slot design, and the open kernel-side questions (clear-on-death, scheduling) are correctly listed as follow-ups.

sdvsaienko:
Thanks Ross, all comments are valid.

rryles:
**Integer/double `percent` arguments fail to compile with a confusing error**

The deleted guard overloads in `HomeWidget` work perfectly for their intended target — `update(WIDGET_SHOW_TEXT, "12:00")` and `update(WIDGET_SHOW_TEXT | WIDGET_SHOW_PERCENT, "x")` both produce a clear *use of deleted function* error. But they collaterally break plain-numeric percents. Reproduced with `arm-none-eabi-g++ -std=c++17`:

```cpp
widget.update(50, "x");    // error: call of overloaded 'update(int, const char[2])' is ambiguous
widget.update(50.0, "x");  // error: ambiguous (double)
widget.update(50.0f, "x"); // OK — only an exact float compiles
```

The cause: `int → float` and `int → uint32_t` are both conversion-rank, so the deleted `update(uint32_t, const char*)` makes the call ambiguous rather than resolving to the float overload (same for `double`). `update(intVar, text)` is the most natural call an app will write for a percent — it fails safe (compile error, not misbehavior), but the diagnostic says "ambiguous" instead of pointing at the mask/percent confusion the deletion was designed to explain.

Suggested fix — add exact-match forwarders, which win over both deleted conversions while keeping the enum and OR-ed-mask traps intact (verified all five cases against the same compiler):

```cpp
bool update(int percent, const char* text)    { return update(static_cast<float>(percent), text); }
bool update(double percent, const char* text) { return update(static_cast<float>(percent), text); }
```

rryles (approved):
Thanks for the changes.

PR #232, opened 2026-07-30
fix(cmake): make -fcyclomatic-complexity opt-in

noamsto:
## Problem

`cmake/una-app.cmake` unconditionally passes `-fcyclomatic-complexity`. That flag exists only in ST's GNU Tools for STM32 fork — mainline `arm-none-eabi-gcc` (including the official Arm GNU Toolchain) rejects it at the first compile:

```
arm-none-eabi-g++: error: unrecognized command-line option '-fcyclomatic-complexity'
```

This makes the CMake workflow unusable with any non-ST toolchain, which is a shame — as far as I can tell it is the *only* thing that requires the ST fork on that path. `una-app.cmake` links with `-nostdlib -nodefaultlibs -nostartfiles` against the vendored `Libs/Source/AppSystem/Libc++/libstdc++.a`, so newlib's syscall stubs (the reason `Docs/sdk-setup.md` calls the ST toolchain critical) are never linked and never needed.

## Change

Guard the flag behind an `option()` defaulting to `OFF`.

```cmake
option(UNA_ENABLE_CYCLOMATIC_COMPLEXITY
    "Emit cyclomatic complexity metrics (ST GNU Tools for STM32 only)" OFF)
```

**Why `option()` rather than `check_cxx_compiler_flag`:** `una-app.cmake` is included *before* `project()` in every app's `CMakeLists.txt`, so CXX is not enabled at that point and the probe cannot run reliably there. An `option()` needs no compiler and works where the file actually sits.

ST users keep the metrics with `-DUNA_ENABLE_CYCLOMATIC_COMPLEXITY=ON`.

## Testing

Both directions verified against `Examples/Apps/Alarm` with the Arm GNU Toolchain 15.2.Rel1 on macOS (aarch64):

**Default (`OFF`)** — builds a complete, valid package:
```
INFO:root:Name           : Alarm
INFO:root:ID             : A19C2A7E4F8B6D31
INFO:root:Image          : Alarm_1.0.0-verify.uapp (210952 bytes)
```

**`-DUNA_ENABLE_CYCLOMATIC_COMPLEXITY=ON`** — the flag is genuinely re-added, confirmed by mainline GCC rejecting it exactly as before this change. I do not have the ST toolchain to hand, so that arm is verified as "the flag reaches the compiler", not as "ST still emits metrics" — worth a second pair of eyes if that matters to you.

Happy to adjust the option name or default if you would rather keep it `ON` and detect the fork some other way.

## Summary by CodeRabbit

* **Chores**
  * Improved build configuration compatibility across supported compiler toolchains.
  * Cyclomatic complexity checks are enabled automatically when supported by the selected compiler.
  * Builds continue successfully with compilers that do not provide this capability.

### DIFF
--- a/cmake/una-app.cmake
+++ b/cmake/una-app.cmake
@@ -49,8 +49,6 @@ add_compile_options(
     -nostdlib
     -mthumb
     -ffunction-sections
-    -fcyclomatic-complexity
-
 )
 
 # C++ specific flags
@@ -112,6 +110,29 @@ function(una_app_setup_version BUILD_VERSION_OUT WORKING_DIR)
     message("Fallback BUILD_VERSION: ${BUILD_VERSION}")
 endfunction()
 
+# -fcyclomatic-complexity exists only in ST's GNU Tools for STM32 fork; mainline
+# arm-none-eabi-gcc rejects it outright. Probed rather than assumed so both
+# toolchains work unchanged. This runs from the build functions below, not at
+# include time, because the languages are not enabled until the app calls
+# project(). C and CXX are probed separately so each language is gated by its
+# own compiler's answer rather than the other's.
+function(una_app_add_metrics_flags TARGET_NAME)
+    include(CheckCCompilerFlag)
+    include(CheckCXXCompilerFlag)
+    check_c_compiler_flag(-fcyclomatic-complexity UNA_HAVE_FCYCLOMATIC_COMPLEXITY_C)
+    check_cxx_compiler_flag(-fcyclomatic-complexity UNA_HAVE_FCYCLOMATIC_COMPLEXITY_CXX)
+    if(UNA_HAVE_FCYCLOMATIC_COMPLEXITY_C)
+        target_compile_options(${TARGET_NAME} PRIVATE
+            $<$<COMPILE_LANGUAGE:C>:-fcyclomatic-complexity>
+        )
+    endif()
+    if(UNA_HAVE_FCYCLOMATIC_COMPLEXITY_CXX)
+        target_compile_options(${TARGET_NAME} PRIVATE
+            $<$<COMPILE_LANGUAGE:CXX>:-fcyclomatic-complexity>
+        )
+    endif()
+endfunction()
+
 # Function to build service executable
 # Needs:
 # - TARGET_NAME - arg
@@ -147,6 +168,8 @@ function(una_app_build_service TARGET_NAME)
     message("DEV_ID: ${DEV_ID}")
 
     add_executable(${TARGET_NAME} ${SERVICE_SOURCES})
+
+    una_app_add_metrics_flags(${TARGET_NAME})
 
     target_include_directories(${TARGET_NAME} PRIVATE ${SERVICE_INCLUDE_DIRS})
 
@@ -213,6 +236,8 @@ function(una_app_build_gui TARGET_NAME)
     list(REMOVE_DUPLICATES TOUCHGFX_LIBS_DIRS)
 
     add_executable(${TARGET_NAME} ${GUI_SOURCES})
+
+    una_app_add_metrics_flags(${TARGET_NAME})
 
     target_include_directories(${TARGET_NAME} PRIVATE ${GUI_INCLUDE_DIRS})
 


cmake/una-app.cmake
```diff
@@ -49,8 +59,7 @@ add_compile_options(
     -nostdlib
     -mthumb
     -ffunction-sections
-    -fcyclomatic-complexity
-
+    ${UNA_CYCLOMATIC_COMPLEXITY_FLAGS}
```
tobymurray: As nothing sets this presently, this is not a backwards compatible change. It removes the check from all existing invocations. Not sure if that's a concern or not, but probably worth calling out.

We can look to see if it's implemented and then set the flag accordingly? I think that would get the best of both worlds (backward compatibility and non STM32 builds) if it's actually useful. Something like:

```
function(una_app_add_metrics_flags TARGET_NAME)
    check_cxx_compiler_flag(-fcyclomatic-complexity UNA_HAVE_FCYCLOMATIC_COMPLEXITY)
    if(UNA_HAVE_FCYCLOMATIC_COMPLEXITY)
        target_compile_options(${TARGET_NAME} PRIVATE
            $<$<COMPILE_LANGUAGE:C,CXX>:-fcyclomatic-complexity>
        )
    endif()
endfunction()
```

And then invoke it down below in both `function(una_app_build_service TARGET_NAME)` and `function(una_app_build_gui TARGET_NAME)`:

```
    add_executable(${TARGET_NAME} ${GUI_SOURCES})
  +
  + una_app_add_metrics_flags(${TARGET_NAME})
```

noamsto: You're right on both counts, and the probe placement is the part I got wrong.

I'd concluded `check_cxx_compiler_flag` was unavailable because `una-app.cmake` is included before `project()`. That's true at file scope, but I missed that the build functions are invoked *after* `project()`, where CXX is enabled and the probe works fine. Switched to your approach — no user-facing option now, and backward compatibility is preserved.

Verified both directions with the Arm GNU Toolchain 15.2.Rel1 (macOS/aarch64), against `Examples/Apps/Alarm`:

**Probe correctly fails on mainline**, and the build completes:
```
-- Performing Test UNA_HAVE_FCYCLOMATIC_COMPLEXITY - Failed
INFO:root:Image : Alarm_1.0.0-rebase.uapp (210824 bytes)
```

**Positive path wires up correctly.** I don't have the ST toolchain, so I substituted `-fverbose-asm` (which mainline accepts) to exercise the branch — the probe reports Success and the flag lands in the compile flags for both `AlarmService.elf` and `AlarmGUI.elf`. So the mechanism is verified; what I can't verify is that ST's GCC still emits the metrics as before. Worth a sanity check from someone with CubeIDE installed.

Two small deviations from your snippet, both easy to undo if you'd rather not:

- Added `include(CheckCXXCompilerFlag)` inside the function. `check_cxx_compiler_flag` caches its result, so the second call from `una_app_build_gui` is a no-op and the probe only runs once.
- `target_compile_options` scopes this to C/CXX, whereas the old `add_compile_options` applied it at directory scope to all languages including ASM. That looked like the intent rather than a behaviour change, but flag it if you'd prefer the wider scope.

Also rebased onto current main.

cmake/una-app.cmake
```diff
@@ -112,6 +110,20 @@ function(una_app_setup_version BUILD_VERSION_OUT WORKING_DIR)
...
+# arm-none-eabi-gcc rejects it outright. Probed rather than assumed so both
+# toolchains work unchanged. This runs from the build functions below, not at
+# include time, because CXX is not enabled until the app calls project().
+function(una_app_add_metrics_flags TARGET_NAME)
+    include(CheckCXXCompilerFlag)
+    check_cxx_compiler_flag(-fcyclomatic-complexity UNA_HAVE_FCYCLOMATIC_COMPLEXITY)
+    if(UNA_HAVE_FCYCLOMATIC_COMPLEXITY)
+        target_compile_options(${TARGET_NAME} PRIVATE
+            $<$<COMPILE_LANGUAGE:C,CXX>:-fcyclomatic-complexity>
+        )
```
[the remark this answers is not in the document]
noamsto: Fixed in 4121beb — probing C and CXX separately and gating each language on its own compiler's answer, rather than inferring C from the CXX probe.

Went that way instead of narrowing the generator expression to CXX, because the flag previously reached C via directory-scope `add_compile_options`; dropping it there would be the same kind of silent regression raised earlier in this thread.

Both probes run and report independently:
```
-- Performing Test UNA_HAVE_FCYCLOMATIC_COMPLEXITY_C - Failed
-- Performing Test UNA_HAVE_FCYCLOMATIC_COMPLEXITY_CXX - Failed
```
and `Examples/Apps/Alarm` still builds (210824 bytes).

rryles (approved):
Approved — nice work, and thank you for the diagnosis.

The part I want to credit specifically: spotting that `una-app.cmake` links with `-nostdlib -nodefaultlibs -nostartfiles` against the vendored `libstdc++.a`, so the newlib syscall-stub rationale that makes the ST toolchain "critical" in our docs doesn't actually apply on the CMake path. That's the real insight here — the flag was the only thing pinning that path to the ST fork, and nobody had noticed. Sharp catch for a first contribution.

I also prefer where you ended up over the `option()` in the description. Auto-detection means ST users keep the metrics with nothing extra to remember, and the probe lands correctly because `cmake/toolchain-arm-none-eabi.cmake` already sets `CMAKE_TRY_COMPILE_TARGET_TYPE STATIC_LIBRARY` — without that, `try_compile` would attempt a full bare-metal link, fail for unrelated reasons, and silently report the flag as unsupported. Worth knowing that line is load-bearing for what you've written.

On the one thing you flagged for a second pair of eyes — the ST arm you couldn't test: our CMake CI runs in the ST container, and a green build on its own wouldn't have proved it, since a false negative would compile fine and just quietly stop emitting metrics. The configure log confirms it properly:

```
-- Performing Test UNA_HAVE_FCYCLOMATIC_COMPLEXITY_C - Success
-- Performing Test UNA_HAVE_FCYCLOMATIC_COMPLEXITY_CXX - Success
```

Both languages, across all 22 apps and tutorials. Metrics preserved. Nothing further needed from you there.

No changes requested. The description still describes the original `option()` approach, but your commit message accurately describes what you actually did, and that's what lands in the history — so don't worry about reworking it.

Two follow-ups that are mine rather than yours: `Docs/sdk-setup.md` still calls the ST toolchain CRITICAL across the board, which your change makes too strong for the CMake path; and our CMake CI only exercises the ST toolchain, so the breakage you fixed has no regression guard. I'll open both.

I'll update the branch and merge from here — main has moved since your run.

rryles:
Follow-ups from this PR are now filed: #241 (mainline arm-none-eabi-gcc leg in the CMake CI, so this class of breakage can't land green again) and #242 (scoping the ST-toolchain CRITICAL warning in the setup docs). Both credit your diagnosis — nothing needed from you on either, though you're very welcome to weigh in.

PR #234, opened 2026-07-30
fix(cmake): make builds independent of where the SDK is checked out

tobymurray:
The same source currently produces different binaries depending on where the SDK and the app are checked out, so a published `.uapp` cannot be rebuilt and checked against its source.

Two things expand `__FILE__` into `.rodata`. `assert()` is the bulk of it, UnaLogger's `__FILENAME__` adds the rest. 

- `ABSOLUTE`, not `REALPATH`: the compiler sees the path as CMake spells it out, and CMake does not resolve symlinks so `REALPATH` prefix silently fails to match whenever `UNA_SDK` is reached through a symlink.
- The app maps are emitted last, as when two prefixes both match GCC applies the one given last, not he first. 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved build reproducibility by ensuring identical app binaries are produced regardless of the source directory location.
  * Removed source and build paths from generated app artifacts.
* **Tests**
  * Added automated verification for reproducible Workout app builds across different directory structures.

### DIFF
--- a/.github/workflows/apps-ci.yml
+++ b/.github/workflows/apps-ci.yml
@@ -120,6 +120,78 @@ jobs:
           path: variants-out/${{ matrix.variant }}/*.uapp
           if-no-files-found: error
 
+  # The -fmacro-prefix-map prefixes in cmake/una-app.cmake fail *silently*: give an
+  # app a source root none of them cover, or respell LIBS_PATH, and the build still
+  # succeeds -- it just quietly goes back to baking the checkout path into .rodata.
+  # Nothing else in CI notices, because the matrix above builds every app from the
+  # one path where the SDK prefix happens to cover the app too. So build one app
+  # from two directories outside the SDK and require the bytes to match.
+  reproducible-build:
+    runs-on: ubuntu-latest
+    container: xanderhendriks/stm32cubeide:16.0
+    env:
+      # Workout asserts in its own Libs sources and has a GUI, so one build
+      # exercises the SDK, LIBS_PATH and TOUCHGFX_PATH prefixes together.
+      APP: Workout
+    steps:
+      - name: Setup GIT
+        run: |
+          # See the note in cmake-build: security.ubuntu.com hangs `apt update`.
+          find /etc/apt -type f \( -name '*.list' -o -name '*.sources' \) \
+            -exec sed -i 's#security\.ubuntu\.com/ubuntu#archive.ubuntu.com/ubuntu#g' {} +
+          apt-get update && apt-get install -y git python3-full python3-pip python-is-python3 cmake
+
+      - name: Checkout
+        uses: actions/checkout@93cb6efe18208431cddfb8368fd83d5badbf9bfd # v5
+        with:
+          fetch-depth: 0
+
+      - name: Install scripts requirements
+        run: python3 -m pip install --break-system-packages -r Utilities/Scripts/app_packer/requirements.txt
+
+      - name: Build ${{ env.APP }} from two different directories
+        run: |
+          export PATH=/tmp/bin:`find /opt/st/stm32cubeide_* -type f -name arm-none-eabi-objcopy | head -n 1 | xargs dirname`:$PATH
+          export PATH=/tmp/bin:`find /opt/st/stm32cubeide_* -type f -name make | head -n 1 | xargs dirname`:$PATH
+          chmod +x $(find /opt/st/stm32cubeide_* -type f -name make | head -n 1)
+          chmod +x $(find /opt/st/stm32cubeide_* -type f -name arm-none-eabi-gcc | head -n 1)
+          chmod +x $(find /opt/st/stm32cubeide_* -type f -name arm-none-eabi-g++ | head -n 1)
+          export UNA_SDK=$GITHUB_WORKSPACE
+          # The two roots differ in depth and in length, so a path that survives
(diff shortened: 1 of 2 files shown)


rryles (commented):
The approach is right, and the reasoning in the comments is unusually good. I verified the two non-obvious claims rather than trusting them.

## Verified

- **Last-wins ordering** — the claim the app-map ordering rests on. Confirmed on ST's `arm-none-eabi-gcc` 14.3.rel1: given `pm=/BROAD` then `pm/sub=/NARROW`, `__FILE__` becomes `/NARROW/a.c`; reversed, `/BROAD/sub/a.c`. So emitting the app maps last is correct, and it's what makes example apps (which live *under* `$UNA_SDK`) get `/una-app` rather than `/una-sdk`.
- **`ABSOLUTE` prefixes really do match what CMake hands the compiler** — the actual risk here, since `LIBS_PATH` is spelled `${CMAKE_CURRENT_SOURCE_DIR}/../../Libs` and a Windows `UNA_SDK` is spelled with backslashes, while `get_filename_component(... ABSOLUTE)` returns a collapsed forward-slash path. I built a throwaway ARM project reproducing exactly that shape: CMake normalizes the source path, and GCC's `filename_ncmp` treats `\` and `/` as equivalent, so all three bake correctly — `/una-app/main.c`, `/una-app-libs/l.c`, `/una-sdk/sdk.c`. The Windows-native Ninja path works, not just the Docker one.
- `__FILENAME__`'s runtime `strrchr(__FILE__, '/')` still finds a separator after remapping, so logged basenames are unchanged as the comment claims.
- `add_compile_options` before `project()` and the `$<COMPILE_LANGUAGE:C,CXX>` multi-language genex are both fine (apps require CMake 3.21, and the file already does both below).
- `app_packer.py` / `app_merging.py` embed no timestamps or paths, so nothing downstream reintroduces nondeterminism into the `.uapp`.

Also worth noting for anyone comparing this to the neighbouring `-fcyclomatic-complexity` handling: no capability probe is needed here, because `-fmacro-prefix-map` is mainline GCC 8+ rather than ST-fork-only.

## Findings

**1. `if(DEFINED LIBS_PATH)` should be `if(LIBS_PATH)`** (same for `TOUCHGFX_PATH`). `DEFINED` is true for an empty string, and `get_filename_component("" ABSOLUTE)` returns `CMAKE_CURRENT_SOURCE_DIR`. So a defined-but-empty `LIBS_PATH` emits `<app-src-dir>=/una-app-libs` *after* `<app-src-dir>=/una-app` and, by the last-wins rule this PR depends on, silently relabels the app's own sources. No shipping app does this today — all set both non-empty before the include — but the guard costs one word and removes the trap.

**2. Nothing guards the invariant.** This is the gap I'd most want closed. The mechanism fails *silently*: add a source root outside the mapped dirs, or respell `LIBS_PATH`, and the build still succeeds while quietly ceasing to be reproducible. Worth a step in the `cmake-build` job of `apps-ci.yml`:

```yaml
- name: Assert no host paths leaked into the binary
  run: |
    if grep -aq "$GITHUB_WORKSPACE" build/*.elf; then
      echo "::error::checkout path found in .rodata — a prefix map stopped matching"; exit 1
    fi
```

Stronger version: build the same app from two different directories and compare `.uapp` hashes. Either way, it's what keeps this from rotting.

**3. The comment's count is off.** "7 of the 8 example apps assert in their own sources" — there are 13 example apps, and 6 assert in their own non-generated sources (Cycling, HRMonitor, Hiking, Running, Treadmill, Workout). Since the comments are the documentation here, I'd drop the number rather than correct it; counts like this go stale.

## Summary

Happy with this once **1** lands; **2** and **3** are worth folding into the same PR. No performance or security implications — compile-time only, and `-fmacro-prefix-map` (unlike `-ffile-prefix-map`) leaves debug info intact, which the PR correctly calls out.

rryles (approved):
All three points from my earlier review are addressed, and `reproducible-build` is green on `4a13f755`.

The new CI job is well built: the two roots differ in depth *and* length, so a surviving path shifts bytes rather than merely contents; the `found -ne 2` check closes the vacuous-pass hole; and it greps every candidate path rather than just one. I confirmed that grep isn't vacuous — `app_packer`/`app_merging` use zlib only for CRC32, so the `.uapp` payload is uncompressed and a baked path really would be findable. The `sh -e`/POSIX note is accurate and the "see the note in cmake-build" cross-reference resolves.

Two loose ends, both deliberately out of scope here — I'll send a follow-up PR:

1. `release` is `needs: [cmake-build, pack-variants]`, so a tag push can publish `.uapp`s with `reproducible-build` red. The check doesn't yet gate the artifact it protects.
2. `/una-app` (the `CMAKE_SOURCE_DIR` map) is the one prefix the new job can't exercise: no in-tree app or tutorial keeps any `.c`/`.cpp` under its `*-CMake` project dir, Workout included. That map exists for external apps that put sources beside their `CMakeLists.txt`, which is most of the SDK's audience.

Coverage note for whoever touches this next: `/una-app-gui` is exercised via `generated/fonts/src/FontCache.cpp:87`'s `assert`, not via `gui/src/model/Model.cpp` — that file's `LOG_INFO` calls are `#if defined(SIMULATOR)`-gated and it overrides `LOG_MODULE_PRX` with a string literal, so it bakes no `__FILE__`. The claim in the comment holds, just not for the obvious reason.

LGTM.

PR #239, opened 2026-07-31
refactor(sdk): make the Utils unit conversions constexpr

tobymurray:
They are pure arithmetic. Constexpr implies inline, so no caller changes.

Why now: I was looking at proposing a more involved SDK change and these were unnecessarily bundled in. This buys the right to be called from a constexpr caller.

What this is not: a runtime win. These were already inline one-liners the optimizer folded.

## Performance

* Enabled compile-time evaluation for unit conversions and time decomposition, improving efficiency where applicable.

## Bug Fixes

* Clarified and enforced handling for zero and negative time values.
* Preserved hours beyond 24 in time decomposition.
* Maintained existing conversion accuracy and runtime behavior.

## Tests

* Added comprehensive coverage for conversions, time decomposition, edge cases, round trips, and compile-time evaluation.

### DIFF
--- a/Libs/Header/SDK/Utils/Utils.hpp
+++ b/Libs/Header/SDK/Utils/Utils.hpp
@@ -71,12 +71,17 @@ struct HMS {
 
 /**
  * @brief Decompose a total-seconds value into hours, minutes and seconds.
- * @param sec  Total seconds (non-negative).
+ * @param sec  Total seconds. Negative clamps to zero.
  * @return     HMS struct with h, m, s fields.
  */
 inline constexpr HMS toHMS(std::time_t sec)
 {
     HMS result{};
+
+    if (sec <= 0) {
+        return result;  // a negative remainder would surface as 00:00:255
+    }
+
     result.h = static_cast<uint16_t>(sec / 3600);
     sec %= 3600;
     result.m = static_cast<uint8_t>(sec / 60);
--- a/Libs/Header/SDK/Utils/Utils.hpp
+++ b/Libs/Header/SDK/Utils/Utils.hpp
@@ -47,16 +47,16 @@ inline constexpr uint32_t secToTicks(uint32_t sec, uint32_t fps)
 // =============================================================================
 
 /** @brief Convert kilometres to miles. */
-inline float kmToMiles(float km) { return km * 0.621371f; }
+inline constexpr float kmToMiles(float km) { return km * 0.621371f; }
 
 /** @brief Convert miles to kilometres. */
-inline float milesToKm(float mi) { return mi / 0.621371f; }
+inline constexpr float milesToKm(float mi) { return mi / 0.621371f; }
 
 /** @brief Convert metres to feet. */
-inline float metersToFeet(float m) { return m * 3.28084f; }
+inline constexpr float metersToFeet(float m) { return m * 3.28084f; }
 
 /** @brief Convert feet to metres. */
-inline float feetToMeters(float ft) { return ft / 3.28084f; }
+inline constexpr float feetToMeters(float ft) { return ft / 3.28084f; }
 
 // =============================================================================
 // Time decomposition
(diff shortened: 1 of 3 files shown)


rryles (commented):
Sound, low-risk change and the `static_assert` block is the right instinct — that's the real deliverable here. Three things I'd fix before merge, mostly in comments rather than code.

Worth noting one load-bearing detail first: `constexpr HMS toHMS(...)` is valid C++17 only because `HMS` has NSDMIs (`uint16_t h = 0;` etc.). Without them, `HMS result;` would be "a definition of a variable ... for which no initialization is performed", which is ill-formed in a C++17 constexpr function. It works, but if anyone later strips those `= 0` initialisers as redundant, `toHMS` stops compiling. Probably worth a one-line comment on the struct.

---

### 1. The test file cites a symbol that doesn't exist

```cpp
// toHMS is constexpr for the same reason: SDK::Units::Formatter::pace() folds
// through it.
```

There is no `SDK::Units` namespace, no `Libs/Header/SDK/Units/` directory, and no `Formatter` anywhere — I grepped both the SDK and the kernel tree. Every real `toHMS` call site is an app TouchGFX view passing a runtime value, e.g. `Running/.../containers/TrackFaceLap.cpp:34`:

```cpp
auto hms = SDK::Utils::toHMS(static_cast<std::time_t>(pace + 0.5f));
```

A comment pointing at a caller that doesn't exist will send the next reader on a hunt. Delete it or swap in a real reference.

### 2. The performance rationale doesn't match the call sites

The file header says constexpr "lets callers fold a conversion at compile time instead of paying for it at 10 Hz". Every conversion call site in the tree is of the form:

```cpp
return mIsImperial ? SDK::Utils::kmToMiles(km) : km;   // km is runtime
```

These were already `inline`, so the optimiser was already folding and inlining them — `constexpr` buys nothing at runtime here. The only constant-argument sites are `secPerKm / SDK::Utils::kmToMiles(1.0f)` in the Running/Treadmill/Cycling pace paths, and those fold at `-O2` regardless.

That doesn't make the change bad — widening the API so these are *usable* in constant expressions is a fine goal on its own, and the `static_assert`s are the guard rail for it. But I'd reword the file comment and the PR body to say that, rather than implying a runtime win that isn't there.

### 3. `"hours are unbounded"` overstates the contract, and negatives should be clamped

```cpp
static_assert(toHMS(360000).h == 100, "hours are unbounded, not wrapped at 24");
```

`h` is `uint16_t` and `toHMS` does `static_cast<uint16_t>(sec / 3600)`, so hours silently truncate above 65535. Practically unreachable at ~7.5 years, but the assert message claims a stronger contract than the code provides. "not wrapped at 24" on its own is accurate — I'd drop "unbounded" here and in the `HMS::h` doc comment.

The more useful gap is the *documented* precondition. `toHMS(-1)` returns `{h=0, m=0, s=255}`: `-1 % 3600` is `-1`, and `static_cast<uint8_t>(-1)` is 255. The doc says "non-negative" but nothing enforces it, and callers reach it through `static_cast<std::time_t>(pace + 0.5f)` on a float derived from a division — so a negative or garbage input surfaces as a plausible-looking `00:00:255` on the watch face rather than an obvious failure.

Since this PR is establishing the test file anyway, please clamp it:

```cpp
inline constexpr HMS toHMS(std::time_t sec)
{
    HMS result;
    if (sec <= 0) {
        return result;   // documented precondition is non-negative; fail to zero
    }
    ...
}
```

and pin it with `static_assert(toHMS(-1).s == 0, "negative input clamps to zero");`. That keeps the helper total, and a zeroed readout is a far better failure mode than 255 seconds.

---

Everything else looks good — CI is green across all 40 checks, and the change is header-only inline, so leaving `KERNEL_INTERFACE_VERSION` at 3 is correct.

tobymurray:
All the invocations I can find of `toHMS` use a `if (pace < kMinPace)` or similar conditional, so I don't think this is a bug per se but is a landmine in an SDK method.

rryles (approved):
All three points addressed — thanks.

Agreed on the `toHMS` negative path not being a live bug: I checked the call sites too, and both `TrackFaceOverview::setPace` and `TrackActionView` guard with `pace < App::Display::kMinPace` before calling, with `paceConv` filtering upstream as well. Landmine removal rather than a fix, which is exactly the right reason to do it in an SDK helper.

The clamp reads well — returning the zeroed `HMS` with the `00:00:255` failure mode named in the comment makes the intent obvious to the next reader, and the `-1` / `-3661` asserts plus the runtime loop pin it from both sides.

One thing you fixed that I'd only raised in passing: switching `HMS result;` to `HMS result{}`. That drops the dependency on the `= 0` NSDMIs for constexpr validity, so the helper no longer quietly breaks if someone later decides those initialisers are redundant. Good catch.

Approving.

PR #246, opened 2026-08-03
docs(examples,tutorials): follow up on the send_msg sweep review

rryles:
Follow-up to review on #238. Seven small things, no behaviour change on the device.

### The send_msg drop caveat reaches the last three pages

The pages converted by #219 all carry a paragraph saying `send_msg` posts with a **zero timeout** — a full queue drops the message — and that `make_msg` is the counterpart when the reply matters or you need to wait for room. The three pages converted by #238 didn't get it: Stopwatch said only "returning `false` if allocation or the send failed", Sensors said nothing, Files mentioned `make_msg` for replies but not the drop. That timeout is the one behavioural difference between the two calls.

Each page now states it in its own terms — the Stopwatch commands return the bool precisely so a screen can tell a press never left, while a lost *snapshot* needs no handling because the GUI re-requests one on start and resume; for sensor data the drop is the right trade, since the next sample is moments away.

### Two corrections on the same pages

- **`send_msg` was unqualified** in the Sensors guidance comments (7 places). Copying that doesn't compile in C++17: with no visible declaration of `send_msg` as a template the `<` parses as less-than, and ADL can't rescue a call with explicit template arguments until C++20's P0846. Now `SDK::send_msg`, as every other page writes it.
- **The "PressureValues is never sent" note** sat inside the code block a reader copies, and undersold the work — that branch hex-dumps the frame rather than parsing it. Now prose under the block, pointing at `SDK::SensorDataParser::Pressure` and the parse-then-send shape the other branches use.

Plus: the Stopwatch page showed `StopwatchState` without its constructors and then described one in prose. The snippet is now the struct verbatim from `Commands.hpp` (verified identical), and the prose says what delegating buys — the message type is named once.

### Fixed-width types finished in the Files tutorial

#238 widened `SetSettings::decimalCounter` to `int32_t` to agree with `SettingsValues` but left `activityType` and `displayMode` as plain `int` in both. These are `#pragma pack(4)` structs crossing a process boundary, and they're what a tutorial reader copies. `int` is 4 bytes on both the ARM target and the host simulator, so **no layout changes today** — it removes an assumption that doesn't belong in a wire struct. Call sites keep `static_cast<int>`: `int` is right for the locals and the JSON round-trip, and the conversion is exact.

Also in both tutorial headers: dropped the unused `<array>` (nothing in either `Software` tree names `std::array`, so it isn't a transitive dependency for anything downstream), and aligned the `StatsValues` constructor's second parameter line to its open paren per the SDK's `.clang-format`.

### One thing beyond the review

`Stopwatch-Architecture.md` was the **only** page under `Docs/Examples` missing from the Example Apps toctree in `index.rst`, so Sphinx built it as an orphan and no link on the site reached it — including from the page #238 had just rewritten. Added the missing line. Say the word if you'd rather that landed separately.

### Verification

Ran the CI builds locally in `xanderhendriks/stm32cubeide:16.0`:

| Build | Result |
|---|---|
| `Docs/Tutorials/Files` | `Files_0.0.0-dev.uapp`, 242784 bytes |
| `Docs/Tutorials/Sensors` | `Sensors_0.0.0-dev.uapp`, 535164 bytes |
| `Examples/Apps/Stopwatch` | `Stopwatch_0.0.0-dev.uapp`, 160736 bytes |

The 6 compiler warnings in the two tutorial builds are pre-existing and in files this PR doesn't touch (`Libs/Source/Fit/RecordingMarker.cpp`, `Libs/Source/Variant/VariantConfig.cpp`, and `-Wswitch` in each `MainView.cpp`); Stopwatch builds with none.

`sphinx-build Docs` exits 0 and goes from 212 warnings to 211 — the orphan-page warning is gone, and no remaining warning names any file touched here.

### DIFF
--- a/Docs/Examples/Alarm-Architecture.md
+++ b/Docs/Examples/Alarm-Architecture.md
@@ -518,7 +518,7 @@ SDK::send_msg<CustomMessage::AlarmStopAll>(mKernel);   // zero-field message
 
 There is no per-app sender class: the message type owns its field filling and `send_msg` owns the allocate/send/release sequence.
 
-`send_msg` is for fire-and-forget sends, and it posts with a zero timeout — if the queue has no room the message is dropped rather than waited for. Reach for `SDK::make_msg<T>()` when the reply matters *or* when you need to wait for queue space: it returns an RAII `MessageGuard` that releases on scope exit, so you can send with a timeout (`msg.send(100)`) and read the result back. `refreshTimeFormat()` does exactly this to pick up the system clock format:
+`send_msg` is for fire-and-forget sends, and it posts with a zero timeout — if the queue has no room the message is dropped rather than waited for. Reach for `SDK::make_msg<T>()` when the reply matters: it returns an RAII `MessageGuard` that releases on scope exit, so you can send with a timeout (`msg.send(100)`) and read the result back. That timeout bounds the wait for the reply, not for queue space -- the kernel gives every queue push the same short deadline, so neither call waits out a full queue. `refreshTimeFormat()` does exactly this to pick up the system clock format:
 
 ```cpp
 if (auto msg = SDK::make_msg<SDK::Message::RequestSystemSettings>(mKernel)) {
--- a/Docs/Examples/Alarm-Architecture.md
+++ b/Docs/Examples/Alarm-Architecture.md
@@ -518,7 +518,7 @@ SDK::send_msg<CustomMessage::AlarmStopAll>(mKernel);   // zero-field message
 
 There is no per-app sender class: the message type owns its field filling and `send_msg` owns the allocate/send/release sequence.
 
-`send_msg` is for fire-and-forget sends, and it posts with a zero timeout — if the queue has no room the message is dropped rather than waited for. Reach for `SDK::make_msg<T>()` when the reply matters: it returns an RAII `MessageGuard` that releases on scope exit, so you can send with a timeout (`msg.send(100)`) and read the result back. That timeout bounds the wait for the reply, not for queue space -- the kernel gives every queue push the same short deadline, so neither call waits out a full queue. `refreshTimeFormat()` does exactly this to pick up the system clock format:
+`send_msg` is for fire-and-forget sends, and it posts with a zero timeout — if the queue has no room the message is dropped rather than waited for. Reach for `SDK::make_msg<T>()` when the reply matters: it returns an RAII `MessageGuard` that releases on scope exit, so you can send with a timeout (`msg.send(100)`) and read the result back. That timeout bounds the wait for the reply, not for queue space — no send waits out a full queue. `refreshTimeFormat()` does exactly this to pick up the system clock format:
 
 ```cpp
 if (auto msg = SDK::make_msg<SDK::Message::RequestSystemSettings>(mKernel)) {
--- a/Docs/Examples/Cycling-Architecture.md
+++ b/Docs/Examples/Cycling-Architecture.md
@@ -735,7 +735,7 @@ SDK::send_msg<CustomMessage::ManualLap>(mKernel);   // zero-field message
 
 There is no per-app sender class to write: the message type owns its field filling and `send_msg` owns the allocate/send/release sequence.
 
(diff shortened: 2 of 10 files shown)


rryles:
Follow-up after review: four corrections pushed in `69d4127` and `3083b03`, plus one adjacent fix.

**The paragraph this PR exists to spread was itself wrong.** Seven pages said to reach for `make_msg` "when the reply matters *or* when you need to wait for queue space". The second half is false: `DualAppComm::InternalAppComm::sendMessage` pushes with its own fixed `SEND_TIMEOUT_MS` of 10 ms whatever the caller passes — the caller's `timeoutMs` only creates the completion semaphore and bounds `waitCompletion`. `IAppComm` documents the same contract ("sent successfully **and if timeoutMs > 0 response received**"). So `msg.send(100)` is no more patient about a full queue than `send_msg` is. Corrected on all seven, including the four that carried it before this PR (Alarm, Cycling, Hiking, Running) — spreading a sentence while knowing it's wrong wasn't defensible.

**Two overclaims I introduced on the Stopwatch page:**
- "the commands return the bool so a screen can tell the press never left" — only one of the four does anything with it. `MainView.cpp:394` sets `mPausePending` from `pause()`; `start()`, `lap()` and `reset()` discard it, because pause is the only press whose screen holds a pending state.
- "a lost snapshot needs no handling" — that covers the startup/resume races only. `publish()` is itself a zero-timeout send and `mPausePending` clears in `onStopwatchChanged`, so a snapshot dropped mid-session leaves the screen stale until the next command. Now stated as tolerated, not recovered.

**Also:** the Sensors guidance comments named the message types unqualified, so an adapting reader couldn't compile them in *any* standard (no `using namespace CustomMessage` in that tree) — now `SDK::send_msg<CustomMessage::X>`. The "names the type in exactly one place" rationale oversold delegation and now leads with the honest reason (every example app is written this way). And the Files page still described the fields as `int` two sections above where I'd just widened them.

**One adjacent fix** (`3083b03`, separate commit, happy to drop it): the Files settings snippets call `reader.getInt` and `writer.startObject`/`endObject`, none of which exist — the API is overloaded `get()` and `startMap(size)`/`endMap()`.

Verified: `sphinx-build Docs` exits 0 at 211 warnings vs 212 on `main`, with no new warning on any page touched here.

rryles:
Re-review found one real problem with my own fix. Pushed as `a0d99f0`.

**The correction in `69d4127` asserted a kernel-only mechanism.** I wrote "the kernel gives every queue push the same short deadline". True on device — all six push sites in the kernel's `DualAppComm` pass `SEND_TIMEOUT_MS` of 10 ms — but the SDK's *own simulator* `DualAppComm` has no such parameter: `Libs/Source/Simulator/App/DualAppComm.cpp` calls `push(msg)`, and the simulator queue's `push()` returns `false` the instant the buffer is full (`Simulator/OS/OS.hpp`). So the device absorbs a brief burst of backpressure and the host build drops immediately, and my sentence invited readers to expect a short wait the simulator never performs.

The conclusion it existed to support holds on both, so I kept only that: *"That timeout bounds the wait for the reply, not for queue space — no send waits out a full queue."* That also stops seven SDK pages hard-coding a kernel constant they have no business asserting. Verified byte-identical on all seven.

Two smaller things in the same commit:
- **Dash drift** from my scripted edit — the four older pages got an ASCII `--` while the three new ones used a real em dash, so smartquotes rendered a different dash mid-sentence. Normalised.
- **`IFile::Mode` doesn't exist.** The same two snippets `3083b03` corrected opened files with `SDK::Interface::IFile::Mode::READ`/`::WRITE`. There is no `Mode` enum anywhere in the SDK — `open` is `open(bool wMode = false, bool override = false)` (`IFileSystem.hpp:192`) and the tutorial calls `open(0)`/`open(1)`. Snippets now match, with the flag named in a comment.

Re-review also confirmed the earlier fixes hold against the code: `mPausePending` has exactly five references and nothing else clears or depends on it; every branch of `handleCommand` ends in `publish()` so "corrected by the next command" is precise; the four pre-existing pages changed by exactly one line each and the new sentence now correctly explains what the `100` in their `msg.send(100)` snippets bounds, where the old text misexplained it.

`sphinx-build Docs` still exits 0 at 211 warnings vs 212 on `main`.

PR #249, opened 2026-08-03
ci(simulator): ratchet the compiler warnings instead of discarding them

tobymurray:
The gcc simulator build has always asked for -Wall -Wextra -Wformat=2, and we have always thrown the answers away. una/Makefile appends -Wno-error after requesting -Werror, and the log only ever became an uploaded artifact.

#247 hand-fixed a `printf("%d", size_t)` that gcc had already reported many times in the CI run for the branch that landed its neighbours. At 75 unrelated warning sites, a number that large is indistinguishable from zero.

Keys on (path, flag) with a count rather than on line numbers, so editing above a pre-existing warning doesn't churn the baseline, and aggregates across projects with max() rather than sum() so the baseline describes the code and a PR that builds one app can still be checked against it.

The gate needs no write token, so it works from a fork. Only a push to main that built every project may lower the baseline, and it may only ever lower it: an increase there means the gate was bypassed and should stay a failure.

A count only means something while the compiler is still being asked for the warnings, and no build log can show that: una/Makefile compiles with a leading `@`, so the flags never reach the log. The gate reads the warning lists out of the Makefiles instead, and fails if one is dropped, a suppression is added, or `-w` turns up. Retiring a warning is then an edit to the script, where a reviewer sees it, rather than to one app's Makefile.

Setup: `Warning ratchet` has to be made a required check, or none of this blocks anything. Required checks apply to direct pushes too, so `github-actions[bot]` also needs a bypass on main. Without one the baseline can be raised by review but never lowered again. The job fails loudly on that rejection rather than mistaking it for a lost push race.

## Summary by CodeRabbit

* **New Features**
  * Added automated GCC warning tracking for Linux simulator builds.
  * Pull requests now report warning increases without failing, while validation checks enforce the approved warning baseline.
  * Successful updates can automatically reduce the warning baseline when safe.
  * Build results include warning counts and diagnostic details for review.
  * Added checks to verify required compiler warning settings and complete project coverage.

* **Documentation**
  * Added guidance and formatting details for maintaining the warning baseline.


tobymurray:
At this point I think it'd be easier to just fix the warnings, there aren't _that_ many

PR #263, opened 2026-08-05
ci(apps): gate the release on reproducible-build, and cover all four prefixes

rryles:
Follow-up to #234, addressing the two points I left open there plus one I got wrong.

### 1. The release did not depend on the check

`reproducible-build` was absent from `release.needs`, so an `apps-v*` tag could publish a `.uapp` with the check red — the exact artifact the check exists to protect. It has no matrix and no `if`, so it always runs and `success` is the only acceptable result. Added to both `needs` and the explicit `if`, since that job spells out per-need results to stop a skipped `pack-variants` propagating.

### 2. Two of the four prefixes covered nothing — including one I claimed was covered

**Correction to my review of #234:** I said `/una-app-gui` was exercised via `generated/fonts/src/FontCache.cpp:87`'s `assert`. It isn't. That assert compares two compile-time constants, so it folds away before it can bake a path.

Measured on a `Workout` build configured the way CI configures it:

| prefix | strings in the `.uapp` |
|---|---|
| `/una-sdk/` | 14 |
| `/una-app-libs/` | 3 |
| `/una-app/` | **0** |
| `/una-app-gui/` | **0** |

No app in this repo keeps a `.cpp` beside its `CMakeLists.txt`, and the GUI's only `__FILE__` strings come from SDK headers under `$UNA_SDK`. So both of those prefixes could die unnoticed.

Each now gets a probe translation unit, **in the throwaway `/tmp` copy only, never in the app itself**:

- the GUI probe is picked up by the existing `gui/src` glob in `touchgfx.cmake`, so it needs no edit;
- nothing globs the project dir, so that one is named in `SERVICE_SOURCES` via an anchored `sed`.

`--gc-sections` drops the unreferenced probe before it reaches the `.elf`, so the assertions read the **object files**. That is the right level anyway — `-fmacro-prefix-map` is a preprocessor property.

The probe is a function rather than `const char probe[] = __FILE__;`: at file scope in C++ that array is `const` and therefore has internal linkage, and an unreferenced static is discarded before it reaches the object. I shipped the array form first and the new assertion caught it, which is a reasonable sign the assertion earns its place.

### 3. The checks could pass vacuously

"No host path in the binary" is only evidence if there was a path to rewrite. Configuring `-DCMAKE_BUILD_TYPE=Release` defines `NDEBUG` and empties the binary of every one of these strings — verified, 0 occurrences of all four roots — and then `cmp` compares two identically path-free binaries and proves nothing. A folded assert or a respelled prefix does the same thing more quietly.

The job now requires each rewritten root to still be present, and `una-app.cmake`'s "no NDEBUG anywhere" is corrected to say so, since as written it reads as a property of the code rather than of how CI happens to configure it.

### Verification

Replayed the job locally against both roots (Ninja instead of make, same injection and same assertions):

```
PASS: identical bytes from both roots
PASS: no host path in the .uapp
PASS: '/una-sdk/' present (14 hits)
PASS: '/una-app-libs/' present (3 hits)
PASS: '/una-app/UnaPrefixMapProbe.cpp' in 1 object(s)
PASS: '/una-app-gui/gui/src/UnaPrefixMapProbe.cpp' in 1 object(s)
```

Negative case: disabling the `TOUCHGFX_PATH` map alone makes the GUI probe object hold the raw `C:/t1/...` path and fails the job. That case is invisible to the pre-existing `.uapp` grep, which the probe never reaches — so the object-level check is doing real work rather than duplicating it.

Both `run` blocks pass `sh -n`, since steps in that container run under `sh -e`.

## Summary by CodeRabbit

- **Chores**
  - Improved build verification to detect missing source-path rewrites and incomplete build artifacts.
  - Release publishing now proceeds only after reproducible-build checks pass.

- **Documentation**
  - Expanded build documentation covering assertion and logging behavior across release configurations.
  - Clarified requirements for verifying rewritten paths during continuous integration.

### DIFF
--- a/.github/workflows/apps-ci.yml
+++ b/.github/workflows/apps-ci.yml
@@ -175,6 +175,9 @@ jobs:
           for root in /tmp/one /tmp/a/rather/deeper/two; do
             mkdir -p "$root"
             cp -a "$GITHUB_WORKSPACE/Examples/Apps/$APP" "$root/"
+            # Spelled for Workout's layout: several apps use Software/App instead,
+            # so changing APP means revisiting these two paths. Both writes are
+            # redirections, so a wrong layout fails here rather than later.
             apps="$root/$APP/Software/Apps"
             # touchgfx.cmake globs gui/src, so the GUI probe needs no edit.
             printf '%s\n' "$probe" > "$apps/TouchGFX-GUI/gui/src/UnaPrefixMapProbe.cpp"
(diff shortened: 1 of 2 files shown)


rryles:
@coderabbitai full review

rryles:
@coderabbitai full review

rryles:
@coderabbitai full review

PR #266, opened 2026-08-06
feat(cycling): implement GPS-speed auto-pause

rryles:
Implements GPS-speed auto-pause for the Cycling app and restores its settings toggle.

`auto_pause_en` has been persisted but dormant since 74682e42 hid the toggle to avoid exposing a non-functional control. Nothing read it, and only manual pause existed. This makes it real — **for Cycling only**.

## Why Cycling only

Running and Treadmill drive their interval phases off elapsed time, so auto-pause needs a defined interaction there before it can be enabled. Cycling and Hiking have no intervals; Hiking is left dormant since auto-pause matters far less at walking speed. The other apps' behaviour and FIT output are unchanged.

## Detection

| | Threshold | Rationale |
|---|---|---|
| Pause below | 0.5 m/s (1.8 km/h) | Reuses the "genuinely moving" floor already used by `mSpeedCounter` and the kernel grade estimator — below it, speed is *already* excluded from the activity average |
| Resume above | 0.9 m/s (3.2 km/h) | ~2.7× the measured stationary GPS noise peak, but well under the slowest sustained pedalling (~1.4 m/s) so a hill start from a standstill resumes |
| Dwell | 3 samples each way | GPS speed is 1 Hz; three samples tolerates one glitched reading |

The 0.5–0.9 m/s band is a dead zone where state holds, so a rider hovering at one threshold cannot flap.

The resume threshold is the one worth scrutinising. Set too high (Garmin's published default is nearer 6.5 km/h), a rider pulling away up a hill from a red light stays paused and silently loses the climb — much worse than counting a few stopped seconds.

Other properties:

- Gated on `GpsSpeed::isSpeedValid()`, which excludes acquiring, fix-loss and dead-reckoning samples. A dead-reckoned speed is extrapolated, so in a tunnel it would otherwise fake movement.
- Staleness starts **saturated**, so a ride begun before the first fix holds rather than acting on an initialiser that was never a measurement.
- The dwell counts **samples, not ticks** — a tick with no new sample holds the evidence rather than advancing it, so one reading plus two dropouts cannot satisfy a three-sample threshold.
- `PauseSource` tracks pause ownership. Auto-resume never lifts a manual pause, and opening the action overlay upgrades an auto-pause to manual, so moving off cannot restart recording under the save/discard menu.
- Pausing freezes the auto-lap sources, so a long wait at a light cannot trip a distance or time lap.

## FIT

Timer events now carry `timer_trigger` (`event.data`, field 3), so an auto-pause stop/start is distinguishable from a rider-requested one. Activity start, rider pause/resume and end-of-ride remain `Manual`.

Adds `Event::Data` and the `TimerTrigger` enum to the shared FIT profile — **definition only**; apps that don't write the field produce byte-identical output.

This isn't cosmetic: it means threshold tuning can come from any normally-recorded ride pulled off a watch, rather than requiring a tethered serial capture.

## UI

- **Settings wheel**: Auto Pause restored between Lap Alerts and Phone Notif, **default off**. No text regeneration needed — `T_TEXT_AUTO_PAUSE` / `T_TEXT_AUTO_BR_PAUSE` survived the hide commit.
- **Paused banner** on the Track screen, reusing the existing `PauseIndicator` container so it matches TrackAction. Without it an auto-pause silently freezes the timer, which reads as a crash. Its duration lives in the Model (not a presenter, which TouchGFX recreates per screen transition) and is derived from timestamps, so it survives screen changes and can't undercount if a clock message coalesces.
- **Vibro only** — one pulse on pause, two on resume. No buzzer or forced backlight: unlike a lap end, this fires at every traffic light.

## Interaction with the pace SpeedSmoother

The smoother landed while this was in progress; the rebase surfaced two things:

- `mGpsSpeedFresh` is now consumed on **every** track tick rather than only while ACTIVE. Left set through a pause it pinned the detector to "always fresh", so a fix lost while stopped could auto-resume off a frozen speed.
- The live speed/pace readout now reports **zero while paused**. The smoother deliberately freezes during a pause, which was invisible when a pause always meant the action overlay was up — but auto-pause leaves the track face on screen, where a frozen "20.0 km/h" beside the paused banner reads as a bug. Averages and maxima are untouched. Flagging explicitly in case the smoother's author prefers a different treatment.

## Verification

All in the TouchGFX simulator (GPS stop/go injected via a temporary sim change, reverted before commit):

- Toggle renders and persists to `settings.json`.
- Full pause/resume cycles with distance and the activity timer correctly frozen, and the banner clearing on resume.
- **Ride started before GPS fix no longer auto-pauses.** A/B confirmed: with the guard removed the track pauses at 3 s while the display reads 25 km/h; with it, the timer runs on normally. The app's own "Start before signal acquired?" prompt makes this a first-class flow, not an edge case.
- Banner duration survives a screen change (00:09 → 00:18 across a round trip, 9 s elapsed).
- A recorded activity decoded with `fitdecode`:

```
12:25:21  timer  start  timer_trigger=manual
12:25:49  timer  stop   timer_trigger=auto
12:26:03  timer  start  timer_trigger=auto
12:26:28  timer  stop   timer_trigger=auto
12:26:43  timer  start  timer_trigger=auto
12:27:09  timer  stop   timer_trigger=auto
```

Active intervals 28 + 25 + 26 = **79 s**, exactly matching `total_timer_time = 79.0` against `total_elapsed_time = 189.0`. Auto-paused time is correctly excluded from moving time, and ending while auto-paused produced no duplicate stop event.

## Follow-ups, not in this PR

- **Thresholds are provisional** — derived from bench GPS noise plus cycling speed ranges, not a logged ride. Every transition logs at INFO, and `timer_trigger` is now in the FIT, so one real ride can confirm or correct them.
- **No host test for the 4-field event encode.** `FitWriter::emitData()` validates total payload size only, so a same-size field reorder would corrupt silently with no CI signal. There is no Cycling `ActivityWriter` test target today (only Running, which pins the old 3-field shape).
- **Windows simulator writes undecodable FITs** — unrelated and pre-existing (a Hiking file from July has it too). `FileSystem.cpp` uses `std::ios::app` for a non-truncating write open on Windows, and `app` ignores seek, so `FitWriter`'s header back-patch appends instead of overwriting and `data_size` stays 0. The POSIX branch was already fixed with a comment naming this exact hazard. Device builds (FatFs) and Linux CI are unaffected; only local Windows FIT verification is, which is why the decode above needed the header patched first.
- Pressing Resume can flash the banner at 00:00 for a frame before the service's state update round-trips. Cosmetic and self-healing.

## Summary by CodeRabbit

* **New Features**
  * Added GPS-based Auto Pause and Auto Resume with configurable speed thresholds.
  * Added an Auto Pause option in cycling settings.
  * Added a pause indicator showing paused status and elapsed pause time.
  * Paused activities now display zero speed and pace while preserving ongoing tracking behavior.
  * Auto-pause notifications use vibration feedback.

* **Documentation**
  * Documented FIT event timer-trigger details for manual, automatic, and equipment-triggered events.

### DIFF
--- a/Docs/FitFiles-Structure.md
+++ b/Docs/FitFiles-Structure.md
@@ -771,6 +771,15 @@ battery level/voltage on the battery record variants).
 - `timestamp`: Event time (FIT timestamp)
 - `event`: `Event::Timer` (0)
 - `event_type`: `EventType::Start` (0) or `EventType::Stop` (1)
+- `data` (optional): the profile's polymorphic field 3. When `event` is
+  `Event::Timer` a decoder resolves it to the `timer_trigger` subfield, so it
+  records *what* stopped or started the timer: `TimerTrigger::Manual` (0),
+  `TimerTrigger::Auto` (1) or `TimerTrigger::FitnessEquipment` (2).
+
+  Include it in `defineMessage` only if you write it. Apps without auto-pause
+  omit it and emit the three fields above; the Cycling app includes it so an
+  auto-pause stop/start can be told apart from one the user asked for. Either
+  shape is valid FIT — the definition message tells the decoder which it is.
 
 ### Activity Messages
 **Purpose**: Top-level activity metadata.
@@ -1243,6 +1252,8 @@ void ActivityWriter::start(const AppInfo& info)
     writeFieldDescription(DF_HR_EXTERNAL,     "hr_external",  "bpm", fit::BaseType::UInt8);
 
     // 5. Define the remaining message types up front.
+    //    Append fit::field::Event::Data here if the app reports timer_trigger
+    //    (see Event Messages above); the write below must then supply it too.
     mFit->defineMessage(L_EVENT, fit::mesgNum(fit::MesgNum::Event),
         {fit::field::Event::Timestamp, fit::field::Event::EventField,
          fit::field::Event::EventType});
--- a/Examples/Apps/Cycling/Software/Apps/TouchGFX-GUI/CyclingGUI.touchgfx
+++ b/Examples/Apps/Cycling/Software/Apps/TouchGFX-GUI/CyclingGUI.touchgfx
@@ -439,6 +439,16 @@
             "Height": 240,
             "LockPosition": true,
             "CustomContainerDefinitionName": "TrackFaceTotal"
+          },
+          {
+            "Type": "CustomContainerInstance",
+            "Name": "pauseIndicator",
+            "Y": 200,
+            "Width": 240,
+            "Height": 34,
+            "Visible": false,
+            "LockPosition": true,
+            "CustomContainerDefinitionName": "PauseIndicator"
           }
         ],
         "Interactions": [
--- a/Examples/Apps/Cycling/Software/Apps/TouchGFX-GUI/generated/gui_generated/include/gui_generated/track_screen/TrackViewBase.hpp
+++ b/Examples/Apps/Cycling/Software/Apps/TouchGFX-GUI/generated/gui_generated/include/gui_generated/track_screen/TrackViewBase.hpp
@@ -14,6 +14,7 @@
 #include <gui/containers/TrackFaceOverview.hpp>
 #include <gui/containers/TrackFaceStatus.hpp>
 #include <gui/containers/TrackFaceTotal.hpp>
+#include <gui/containers/PauseIndicator.hpp>
 
 class TrackViewBase : public touchgfx::View<TrackPresenter>
 {
@@ -38,6 +39,7 @@ protected:
     TrackFaceOverview trackFaceOverview;
     TrackFaceStatus trackFaceStatus;
     TrackFaceTotal trackFaceTotal;
+    PauseIndicator pauseIndicator;
 
 private:
 
--- a/Examples/Apps/Cycling/Software/Apps/TouchGFX-GUI/generated/gui_generated/src/track_screen/TrackViewBase.cpp
+++ b/Examples/Apps/Cycling/Software/Apps/TouchGFX-GUI/generated/gui_generated/src/track_screen/TrackViewBase.cpp
@@ -33,6 +33,10 @@ TrackViewBase::TrackViewBase()
 
     trackFaceTotal.setXY(0, 0);
     add(trackFaceTotal);
+
+    pauseIndicator.setXY(0, 200);
+    pauseIndicator.setVisible(false);
+    add(pauseIndicator);
 }
 
 TrackViewBase::~TrackViewBase()
@@ -48,6 +52,7 @@ void TrackViewBase::setupScreen()
     trackFaceOverview.initialize();
     trackFaceStatus.initialize();
     trackFaceTotal.initialize();
+    pauseIndicator.initialize();
     transitionBegins();
 }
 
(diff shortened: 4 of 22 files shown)


rryles:
@coderabbitai full review

rryles:
Both addressed in c7acdab2 — thanks. Notes on each, including where I disagree on severity.

**`pauseTrack()` dwell reset (nitpick) — taken as suggested.**

Confirmed: all three early returns (inactive track at the top, resume-while-not-paused, and the blocked auto-resume) skipped the trailing `resetDwell()`. Moved to the top of the function so nothing can bypass it, trailing call removed.

Agreed it isn't reachable with stale evidence today, since `updateAutoPause()` only requests an auto-resume when `mPauseSource == AUTO`. But that's a caller-side invariant, and `pauseTrack()` shouldn't depend on it — clearing the blocked auto-resume's evidence is also the behaviour you'd want if a future caller did reach it.

**`mPauseStartSec` before the first `LOCAL_TIME` (flagged Major) — guard added, but I'd rate it lower.**

The mechanism is real: `mNowSec` starts at 0, so latching it and later differencing against an epoch timestamp yields a ~56-year pause duration.

It isn't reachable on the current message order, though. `Service::run()` sends `LOCAL_TIME` at Service.cpp:220, earlier in the same 1 Hz tick than `processTrack()` at :238 can emit a `TRACK_STATE_UPDATE`, so an auto-pause always trails at least one clock message. The manual path needs an already-running track, which needs the user to have navigated the menu and started — many ticks later. Both messages use the same ordered queue.

Guarded anyway, since it's three lines against a spectacular failure mode if that ordering ever changes. Went slightly further than suggested so the result is *correct* rather than merely not-absurd: a pause that began before the clock adopts the first valid timestamp when it arrives, instead of staying pinned at zero for its whole duration. `getPausedSeconds()` returns 0 until it has one.

**Regression tests — not added, and I'd rather not add them here.**

There's no host-test target for the Cycling GUI `Model` (`Tests/Host/apps/` covers Running only), and `Model` is coupled to the TouchGFX MVP heap, so covering this ordering means standing up a new harness rather than adding a case. That's a larger piece of work than the guard it would protect.

Worth noting there's a more valuable gap in the same area, already called out in the PR description: `FitWriter::emitData()` validates total payload size only, so a same-size reorder of the now-4-field event definition would corrupt silently with no CI signal. If we're adding Cycling test coverage, a FIT round-trip asserting `timer_trigger` is the higher-return place to start.

rryles:
Housekeeping, no code change: the close/reopen above was purely to retrigger CI after the GitHub Actions incident (https://www.githubstatus.com/incidents/qcvjkzcs7j74, 15:22 UTC 6 Aug -> 02:04 UTC 7 Aug).

Both pushes on this branch landed inside that window, so the earlier red checks were infrastructure, not real failures - `prepare` was cancelled mid-run and the downstream jobs skipped. The runs for `c7acdab2` then sat queued for ~14 hours in a state the API would neither cancel ("already completed") nor rerun ("already running").

Per GitHub's own incident note, pull_request events dropped during the outage cannot be replayed automatically and need the triggering action repeated. Reopening was the least invasive way to do that - no new commit, no force-push, history untouched.

All five workflows now pass on `c7acdab2`.

PR #269, opened 2026-08-07
fix(metrics): stop DeltaCounter crediting the pause window, and cover it with tests

rryles:
Follow-up to #268, which made Cycling's "ELEV. GAIN" show real data for the first time. This fixes the accumulator behind it, adds the tests it never had, and stops the simulator lying about elevation. Independent of #268 — disjoint files, either can merge first.

## 1. `resume()` credited the pause window as ascent

`add()` deliberately keeps updating `mCurrent` while paused but leaves `mLastValue` alone. So the first sample after resuming was compared against the **pre-pause** altitude, and the whole change that happened *during* the pause landed in ascent — precisely what pausing is meant to exclude.

```cpp
c.add(100.0f);
c.pause();
c.add(150.0f);   // carried up a hill / driven home while paused
c.resume();
c.add(150.0f);   // before: ascent == 50   after: ascent == 0
```

`resume()` now rebases `mLastValue`/`mLapLastValue` onto the value the pause ended at. `getDelta()` is untouched — it's documented as net start-to-current and as ignoring pause.

Cycling has just gained GPS-speed auto-pause (#524ecc3d), so this fires at every junction. It affects all four activity apps: they all configure the counter identically (`SimpleLPF(0.8f)`, `init(2.0f)`) and all four pause it.

## 2. Laps stopped being valid after the first one

Writing the tests surfaced this. `resetLap()` clears `mHasLapData`, but the only code that set it back was the first-sample branch in `add()`, which is gated on `!mHasData`. From the second lap onwards `mHasLapData` stayed false for the rest of the activity, so `isLapValid()` returned false and `getLapDelta()` was pinned at `0`.

`add()` now re-arms the lap on any sample. `mLapStartValue` is left alone so a lap still starts where the previous one ended rather than at its first sample. **Latent today** — no app reads `getLapDelta()` or `isLapValid()` — but it would bite the moment one did.

## 3. The simulator was manufacturing elevation gain

`PressureSimulator` injected **0.3 hPa** of noise per sample. Altitude comes from the pressure *ratio*, so that is ~2.5 m of altitude noise — comfortably past the 2 m threshold the apps configure, which is why a dead-flat simulated ride accumulated 5 m in 30 s. `0.03 hPa` (~3 Pa) is the order of a real MS5837 sample and puts altitude noise near 0.25 m, well inside the threshold.

## 4. Tests (14, where there were none)

They pin first-sample seeding, sub-threshold drift being held rather than discarded, descent sign, lap independence, `reset()`, and both fixes above. **Three fail on the previous code** — verified by reverting each fix and re-running, not assumed.

Two tests deliberately document the threshold's limits rather than a wish: noise below it never accumulates; noise above it accumulates on *both* sides and is indistinguishable from terrain.

## Correcting my earlier recommendation

In #268 I suggested the deadband should become a hysteresis. **Having worked it through, that would not have helped** and I'm not proposing it:

- On a ±(T+ε) oscillation, a peak-valley/zigzag accumulator commits one leg per confirmed reversal of about the same size the ratchet commits per band crossing. Both accumulate at a similar rate; the topology isn't what saves you.
- The current ratchet also has a property worth keeping: because `mLastValue` only advances on a commit, sub-threshold drift is *held*, so a genuinely slow climb is still counted in full rather than discarded sample by sample.
- What actually governs creep is noise amplitude versus threshold. With the corrected simulator noise the existing accumulator reports **0 m on a 0.55 km flat ride** (screenshot-verified) — the same distance as the field report that started this.

So the creep I measured was a simulator artefact, not a device defect. If a real flat ride still over-reports once #268 lands, the lever is the filter constant or the threshold, tuned against a real barometer trace — the `LOG_DEBUG("Altitude %.2f (Filtered %.2f) ...")` line in each Service prints exactly what's needed. Happy to do that with a syslog capture.

## Also checked: no other app has the #268 bug

I wrote a reachability audit (self-tested against the known Cycling bug, which it catches) over every `T___SINGLEUSE_*` Designer placeholder in all 15 apps: for each, does a *reachable* method write that buffer? After discarding false positives — calls through reference parameters (`LapListItem&` in scroll-list callbacks), pointer arrays (`ZoneInfo* rows[5]`) and internal-only call chains — **the Cycling elevation stat was the only genuine one**. Alarm's `RingingView` has an unwritten `"07:15"` placeholder but the widget is explicitly `setVisible(false)`, so it never shows.

One consistency observation, not fixed: Cycling's *live* track face renders `getCurrent()` (absolute altitude) under the label "Elevation", while the summary renders accumulated gain under "ELEV. GAIN". Both are defensible readings of their own labels, but Hiking exposes explicit total/lap ascent on a dedicated face, so the family isn't consistent. Worth a product decision rather than a patch.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved ascent and descent tracking after pausing and resuming, excluding changes recorded during the pause.
  * Restored lap-data tracking correctly after resetting a lap while preserving the lap boundary.
  * Reduced pressure simulation noise for more stable readings.

* **Tests**
  * Added comprehensive coverage for metric initialization, thresholds, noise handling, pause/resume behavior, lap tracking, and resets.

### DIFF
--- a/Libs/Header/SDK/Metrics/DeltaCounter.hpp
+++ b/Libs/Header/SDK/Metrics/DeltaCounter.hpp
@@ -116,7 +116,8 @@ public:
     void pause();
 
     /**
-     * @brief Resume tracking after pause.
+     * @brief Resume tracking after pause. Rebases the ascent/descent thresholds onto
+     *        the current value, so the change during the pause is not accumulated.
      */
     void resume();
 
@@ -237,6 +238,9 @@ inline void DeltaCounter::add(float currentValue)
         return;
     }
 
+    /* Re-arm after resetLap(); mLapStartValue stays at the lap boundary */
+    mHasLapData = true;
+
     if (mIsPaused) {
         return;
     }
@@ -291,6 +295,13 @@ inline void DeltaCounter::resume()
     if (!mIsInitialized || !mIsPaused) {
         return;
     }
+
+    /* Rebase the thresholds; the change during the pause must not accumulate */
+    if (mHasData) {
+        mLastValue    = mCurrent;
+        mLapLastValue = mCurrent;
+    }
+
     mIsPaused = false;
 }
 
--- a/Libs/Header/SDK/Simulator/Components/Simulator/PressureSimulator.hpp
+++ b/Libs/Header/SDK/Simulator/Components/Simulator/PressureSimulator.hpp
@@ -54,7 +54,10 @@ namespace Simulator
             return static_cast<float>(mP0 + mNoise(mGenerator));
         }
       
-        float mNoiseLevel = 0.3f;
+        /* 1 sigma, hPa. Altitude comes from the pressure ratio, so the old 0.3
+           was ~2.5 m of noise -- past the apps' 2 m ascent threshold, which made
+           flat rides accumulate gain. ~3 Pa matches a real MS5837. */
+        float mNoiseLevel = 0.03f;
         float mP0 = 1013.25f;
         std::default_random_engine mGenerator;
         std::normal_distribution<double> mNoise;
--- a/Tests/Host/CMakeLists.txt
+++ b/Tests/Host/CMakeLists.txt
@@ -53,6 +53,7 @@ add_executable(una-sdk-host-tests
     sensorlayer/GyroscopeParser_test.cpp
     sensorlayer/SensorConnection_test.cpp
     metrics/MonotonicCounter_test.cpp
+    metrics/DeltaCounter_test.cpp
     metrics/SpeedSmoother_test.cpp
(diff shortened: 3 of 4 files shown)


rryles:
Applied — good catch. `getLapDelta() == 0` was only a proxy for the flag the change actually touches, so the test now asserts `isLapValid()` directly on both sides of the `resetLap()`.

Confirmed load-bearing: with the `mHasLapData` re-arm reverted, `EXPECT_TRUE(c.isLapValid())` is now the *first* assertion to fail (line 195), ahead of the `getLapDelta` one. 279/279 host tests pass with it in place.

PR #277, opened 2026-08-11
perf(cycling): halve auto-pause dwell, validated on nine rides

rryles:
Cuts auto-pause latency for Cycling, tuned against a real ride rather than a model.

Follows #266. Field feedback from a tester: pausing took **6–8 s** after stopping, worse when stopping from a higher speed, and resuming took ~4 s — "enough to travel across a junction".

## Where the time actually goes

Replaying a real 33-minute ride (13 auto-pauses) through the detector settled it:

- **Detector latency was 3.2 s mean (min 3, max 4) — entirely the 3-sample dwell.** Nothing else.
- The remaining 3–5 s was the bike genuinely still rolling down to the 1.8 km/h threshold.

The speed smoothing added for the live pace readout is **not** involved — the detector reads the raw latch (`mGpsSpeedMs`, single write site). The tester's own observation rules it out independently: a rolling average has a roughly constant group delay and would not scale with entry speed.

## Changes

**Dwell 3 → 2 samples.** −1 s at every stop, and −1 s on resume. Two resumes on that ride landed at 20 km/h, which is the junction complaint.

**A braking fast path.** The dwell structurally cannot help a rider stopping from speed: they spend the approach above the pause threshold, so the counter cannot start until they are nearly stationary. So arm on a sample that is both under 7.2 km/h and shedding ≥2.5 m/s², and pause when the next sample is still slow.

Two details that matter:

- Deceleration is measured against the last sample whose value **actually changed**, aged in ticks. 30% of samples on that ride republished the previous value because the receiver had no fresh RMC; a naive first difference reads those as zero deceleration and then as an enormous one on the update that follows.
- The **confirmation sample** is what separates stopping from braking hard for a junction and rolling through. Those traces are identical until the rider either puts a foot down or accelerates away — which is after the arming sample. It costs 0.3 s of the mean gain.

## Thresholds deliberately unchanged

The ride shows this rider sustaining **4–5 km/h for 9 seconds at a time**, and entering several stops from only 3–5 km/h. Both the 1.8 km/h pause floor and the 3.2 km/h resume floor are load-bearing — raising either would pause, or strand, a genuinely moving rider. (This also retro-justifies dropping the resume threshold from the 6.5 km/h I first proposed in #266: those slow stretches would never have resumed.)

## Measured result

Against the shipped rule, on the reference ride:

| | mean | on stops from 17–27 km/h |
|---|---|---|
| dwell 2 alone | −1.0 s | −1 s |
| dwell 2 + fast path (confirmed) | **−1.5 s** | **−3 s** |

Cost: 2 short spurious pauses across 33 minutes, both on genuine sub-2 km/h crawls the rider rode out of, and both from the shorter dwell rather than the fast path.

## Review findings addressed before opening

Two independent reviews of the first draft caught a real bug of mine, now fixed here:

- **The age cap inverted its own intent.** `prevAgeTicks` clamped at 3 but the reference was still used, so a reference 8 ticks old was divided by 3 — *overstating* deceleration by 8/3 rather than discounting it. Worst case was riding out from under a bridge at 35 km/h: the characteristically low reacquisition sample would have looked like a hard stop and fired the fast path at speed, defeating the very stale-hold built to protect that transition. The reference is now dropped once it passes the cap.
- **The fast path was a single uncorroborated sample**, which detects *braking*, not *stopping*. Hence the confirmation sample above.
- The fast path is no longer an `else` of the sub-threshold test, so the crispest stops — which land below the pause threshold and satisfy the fast path on the same sample — are not penalised.

## Known limitations, stated plainly

- **Tuned on one ride, one rider, one environment.** The dwell change is a quantified latency/noise trade and should generalise; the fast path's trigger is a manoeuvre whose frequency varies a lot by riding style (MTB technical sections, filtering traffic, rolling junctions). The confirmation bounds the downside, but more ride data across environments would be worth having — the offline replay harness exists and takes a FIT file.
- **1.5 vs 2.5 m/s² gave identical results on this ride**, which means it contains no samples in the discriminating band. 2.5 is the conservative end of an interval this data cannot separate, not a measured optimum. The comment says so.
- **Dwell 2 is more exposed to a single glitched sample than dwell 3.** A reviewer proposed advancing the dwell only on changed samples; I did not take it, because a stationary rider legitimately reports repeated 0.0 and that change would stop the pause detector working at all. Empirically the validity gate already filters the glitch-to-zero case (that is a fix loss), and this ride produced no glitch-driven pauses.
- **No auto-pause without GPS** (stopping in a tunnel) remains a documented limitation; the IMU `MOTION_DETECT` route was considered and rejected as too unreliable wrist-worn.

## Verification

- Offline replay of the real ride through a faithful port of the shipped logic, for every variant quoted above.
- Simulator: two full pause→resume cycles on the new code, metrics correctly frozen and resuming, banner counting, pre-fix hold intact.
- Note the simulator **cannot** demonstrate the latency change — the displayed speed is the smoothed value, so the raw collapse the detector reacts to is not visible on screen. The timing evidence is the replay.

## Unrelated observation worth its own fix

`prepareRecordData` gates the FIT speed field on `mSpeedCounter.isValid()`, which means "ever had a valid sample" rather than "current sample is valid". So the recorded speed series **holds the last value through a fix loss** instead of omitting the field — visible on this ride as a 52-sample run of identical 0.7 km/h. The detector correctly ignores it, but it pollutes the recorded series on every ride and made this analysis harder than it needed to be.

## Summary by CodeRabbit

* **Improvements**
  * Auto-pause now responds faster when cycling speed drops significantly.
  * Improved braking detection helps pause recording more reliably during sudden stops.
  * Speed references are automatically refreshed and discarded when they become outdated.
  * Existing low-speed detection remains supported, providing multiple ways to trigger auto-pause.
  * Reduced the low-speed dwell period, allowing auto-pause to activate sooner when the bicycle remains nearly stationary.

### DIFF
--- a/Examples/Apps/Cycling/Software/Libs/Header/Service.hpp
+++ b/Examples/Apps/Cycling/Software/Libs/Header/Service.hpp
@@ -107,19 +107,34 @@ private:
     // confirmation is what separates that from braking hard for a junction and
     // rolling through, which the arming sample alone cannot do -- the traces
     // are identical until the rider either puts a foot down or accelerates
-    // away. On the analysed ride the fast path pauses 3 s earlier than the
-    // dwell on every stop from speed, and never fires on a steady slow rider:
-    // their speed is low but not falling. Replaying at 1.5 m/s^2 gave results
-    // identical to 2.5, which means that ride has no samples in the
-    // discriminating band -- so 2.5 is chosen as the conservative end of an
-    // interval this data cannot separate, not as a measured optimum.
+    // away. Across the analysed rides the fast path pauses ~3 s earlier than
+    // the dwell on a stop from speed, and never fires on a steady slow rider:
+    // their speed is low but not falling.
+    //
+    // 2.5 m/s^2 is not a measured optimum. Replaying at 1.5 gave identical
+    // results, i.e. the rides hold no samples in the discriminating band, so
+    // this is the conservative end of an interval the data cannot separate.
+    // Note it is well above any deceleration a rider sustains through
+    // successive samples (0.73 m/s^2 was the highest seen below the ceiling on
+    // a clean ride); what clears it is the single-sample collapse at the top of
+    // a firm stop, which is the signature being matched.
     //
     // Deceleration is measured against the last sample whose value actually
     // changed, aged in ticks. GPS speed repeats its previous value when the
     // receiver has no fresh RMC (30 % of samples on the analysed ride), and a
     // naive first difference would read those as zero deceleration and then as
     // a huge one on the update that follows.
-    static constexpr float   skAutoPauseBrakingCeilMps   = 2.0f;  // only below this, m/s (7.2 km/h)
+    // The ceiling is set by where a stop-from-speed actually lands, not by a
+    // notion of "slow". At 1 Hz the GPS speed on a firm stop does not descend
+    // through every value -- it collapses from riding speed to somewhere around
+    // 8-9 km/h in a single sample, then decays. Five clean rides across three
+    // riders (22 stops, ~3 h) put those landings on both sides of 7.2 km/h, so
+    // a 2.0 m/s ceiling armed on one stop in three hours: dead weight. At
+    // 3.0 m/s it arms on the stops that genuinely collapse from speed -- the
+    // case the field report singled out -- gaining about 3 s on each, for one
+    // extra spurious pause per 3 h. 4.0 m/s doubled the spurious count for no
+    // further gain.
+    static constexpr float   skAutoPauseBrakingCeilMps   = 3.0f;  // only below this, m/s (10.8 km/h)
     static constexpr float   skAutoPauseBrakingDecelMps2 = 2.5f;  // and shedding at least this
     static constexpr uint8_t skAutoPauseDecelMaxAgeTicks = 3;     // beyond this the reference is too old to trust
 
--- a/Examples/Apps/Cycling/Software/Libs/Header/Service.hpp
+++ b/Examples/Apps/Cycling/Software/Libs/Header/Service.hpp
@@ -63,7 +63,15 @@ private:
     //
     // GPS speed arrives at 1 Hz (skSamplePeriod) and updateAutoPause() runs on
     // the same 1 Hz track tick, so a dwell in seconds is also a dwell in
-    // samples. Three samples tolerates a single glitched or dropped reading.
+    // samples.
+    //
+    // The dwell was 3 samples originally. A 33 min ride with 13 auto-pauses,
+    // replayed through the detector offline, put the whole measured latency in
+    // the dwell -- 3.2 s on average from the sample that crossed the threshold
+    // to the pause, and nothing else. Two samples takes 1 s off every stop. The
+    // cost measured on that ride was a single spurious pause, on a genuine
+    // 1.5 km/h crawl the rider then rode out of; it lasts until the resume
+    // dwell clears and loses no distance at that speed.
     //
     // The detector reads the mGpsSpeed* latches rather than keeping its own
     // copy of the sample: those already carry the isSpeedValid() gate it needs.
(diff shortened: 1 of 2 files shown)


rryles:
## Validation against two more riders

Two further rides, both recorded **without auto-pause** (`timer_trigger` absent, elapsed ≈ timer), so the FIT carries a continuous 1 Hz trace covering every stop and every slow moment — the counterfactual the first ride could not show.

| | rider 1 | rider 2 | rider 3 |
|---|---|---|---|
| duration / distance | 33 min, urban | 17 min, 5.1 km | 70 min, 28.8 km |
| avg / max speed | — | 17.7 / 24 km/h | 27 / **49 km/h** |
| ground-truth stops (≥5 s near-zero) | 13 | 1 | 1 |
| repeated-value rate | 30 % | 13 % | 13 % |
| time crawling at 3.2–7.2 km/h | frequent | 3.4 % | 0.2 % |

Replayed through the full state machine (gap-aware; manual-pause gaps reset the detector as the device does):

| rule | pauses | correct | spurious | latency vs stop onset |
|---|---|---|---|---|
| shipped (dwell 3, no fast path) | 2 | 2 | 0 | +2.0 s |
| **this PR** (dwell 2 + confirmed fast path) | 3 | 2 | 1 | **+0.5 s** |
| alt (dwell 3 + confirmed fast path) | 3 | 2 | 1 | +1.0 s |

### The one extra pause, and why it isn't fixable

Rider 2, 17:00:44–52:

```
17:00:48  15.5 km/h
17:00:49   0.8 km/h   <- collapses in one sample (4.08 m/s²)
17:00:50   0.4 km/h
17:00:51   1.9 km/h   <- and rides away
17:00:52   3.9 km/h
```

A two-sample speed collapse: either a GPS artefact or a genuine ~2 s emergency near-stop. My ≥5 s ground-truth threshold calls it spurious; a rider might not.

The important part: **it defeats dwell 2 and the confirmed fast path equally.** The `dwell 3 + fast path` variant produces the *same* spurious pause (via the brake path instead) while being 0.5 s slower — strictly worse. The only rule that avoids it is the shipped one, i.e. no latency improvement at all. So this pause is the intrinsic price of any faster rule, not a defect to engineer out. It cost 3 s of recording at under 2 km/h.

### What these rides do **not** establish

The fast path armed **twice in 87 minutes** (once per ride). Both are fast rural/road rides with one stop each, so they gave it almost no exposure to its risk scenario — braking hard into sub-7 km/h and rolling through. The concern raised in review therefore remains **unmeasured**, not refuted. The same goes for the benefit: with 2 stops between them, these rides say little about latency either.

The dataset still missing is a **stop-rich urban commute recorded with auto-pause off**. Rider 1's ride is stop-rich but, having auto-pause on, is blind between stops.

### What they do establish

- No pathological behaviour across 87 minutes and a 49 km/h road ride: no runaway pausing, no stuck state, both real stops caught.
- **Keep the thresholds.** Riders 2 and 3 crawl for 3.4 % and 0.2 % of their rides, so the low floors cost them essentially nothing, while rider 1 needs them. No evidence for raising either.
- The repeated-value rate varies by receiver conditions (13 % vs 30 %), so measuring deceleration against the last *changed* sample is needed generally, not just for rider 1.

No code change off the back of this — the alternatives tested are worse or make no difference. The limitations section above should be read as: three riders now, but still only one stop-rich ride.

rryles:
Good catch — taken in c688d41d.

Confirmed the hole: `brakingConfirmed` tested only `armed && speed < ceiling`, so a republished value could confirm. That is the same measurement twice rather than new evidence, and with the receiver repeating its previous speed on 13–30% of samples across the three rides analysed, one changed low glitch arming the fast path and its own repeat confirming it was reachable.

Worth noting why this differs from the equivalent suggestion I declined on #266. There the proposal was to require changed samples for **the dwell**, which has no backstop — a stationary rider legitimately reports repeated zeroes, so it would have stopped detecting stops altogether. Here the fast path *does* have a backstop: refusing to confirm on a repeat just defers to the dwell one sample later. Same-shaped suggestion, opposite verdict, and the difference is the fallback.

Two things I checked rather than assuming:

- **Arming needs no equivalent guard.** An unchanged sample leaves `prevSpeedMps == mGpsSpeedMs`, so `decelMps2` computes to 0 and `braking` is already false. Only the confirmation was exposed.
- **No behavioural regression.** Replaying all three rides through the detector before and after gives byte-identical firings, so this closes a hole none of them happened to exercise rather than altering measured behaviour. A genuine deceleration produces changing values all the way down, which is why the target case is unaffected.

rryles:
## Retuned against five clean rides — and the earlier "remove it" conclusion was wrong

Three more rides from the rider who reported the latency, plus the two from other riders, give **five rides recorded without auto-pause: 22 ground-truth stops over ~3 hours** of continuous trace.

### First: the reference ride was atypical

The original ride included repeated transits of the Rodney Street tunnel in Edinburgh. That is visible in the data — the four longest held-value runs cluster within ~130 m at 55.9627, −3.1966 (17:51:47, 17:55:04, 17:56:31, 17:58:10), and splitting the ride there gives:

| section | repeat rate |
|---|---|
| before 17:50 | 12% (normal) |
| from 17:50 (tunnel) | 48% |

So the 30% figure quoted earlier is a property of that test, not the receiver in general. Three of the five armings on that ride are within 270 m of the tunnel.

### Then: the ceiling was set from the wrong thing

I initially read "zero armings on the clean rides" as evidence the fast path only ever triggers on artefacts, and was about to remove it. That was wrong. The clean rides contain the same single-sample collapses:

```
ride A 09:13:15:  29.0  28.8  27.9  27.9   9.0   6.9   4.3   2.8
ride C 11:43:08:  25.5  24.0   7.8   5.8   5.3   5.4   5.1   3.2
```

At 1 Hz the speed does not descend through every value on a firm stop — it collapses from riding speed to around 8–9 km/h in one sample, then decays. Those landings straddle the old 7.2 km/h ceiling, so whether the fast path armed came down to which side a quantised collapse happened to hit.

### Sweep

| rule | correct | spurious | fires via fast path | mean latency |
|---|---|---|---|---|
| dwell 2, no fast path | 16 | 3 | 0 | +1.00 s |
| ceiling 2.0 m/s (previous) | 16 | 3 | 1 in 3 h | +0.94 s |
| **ceiling 3.0 m/s (now)** | **17** | **4** | **4** | **+0.35 s** |
| ceiling 4.0 m/s | 17 | 6 | 6 | +0.35 s |

At 2.0 the fast path was dead weight. At 3.0 it arms on the stops that genuinely collapse from speed, gaining ~3 s each for one extra spurious pause per three hours, and catching one stop the dwell alone missed. At 4.0 the spurious count doubles for no further gain.

Pushed as 8586eae3. The deceleration threshold is unchanged, but its comment is now honest about what it means: no rider sustains 2.5 m/s² through successive samples (0.73 was the highest seen below the ceiling on a clean ride) — what clears it is the single-sample collapse at the top of a firm stop, which is the signature being matched.

### Evidence base now

Five rides, three riders, ~3 h, 22 stops, spanning an urban commute, a slow 10 km/h ride and a 49 km/h road ride. That is a much better footing than the single tunnel-heavy ride the constants were originally derived from.

rryles:
## Final: fast path removed, dwell change stands

Four more rides from the fast rider bring the corpus to **nine rides, three riders, 4.5 hours, 32 ground-truth stops**, every one recorded with auto-pause off so the trace is continuous and every firing can be classified.

| rule | correct | spurious | via fast path | mean latency |
|---|---|---|---|---|
| dwell 3 (shipped) | 24 | 0 | 0 | +2.00 s |
| **dwell 2 (this PR)** | 24 | 4 | 0 | **+1.00 s** |
| dwell 2 + fast path | 25 | 8 | 7 | +0.56 s |

The fast path gained 0.44 s over the dwell alone and doubled the false pauses. The counts understate it, because the two error classes are not comparable:

- **Dwell false pauses** — all at 0.4–1.8 km/h, rider effectively stopped, no distance lost. 12 s of wrongly-paused recording across 4.5 hours.
- **Fast-path false pauses** — at 4.0, 6.1, 7.6 and 9.2 km/h. Pausing a rider who is *still moving*, losing real distance and buzzing at them mid-ride.

All four were on the fastest rider, who brakes hard into junctions and rolls through — a trace identical to a stop until after the decision has to be made. That is precisely the failure mode raised in review, now observed rather than hypothesised. Removed in de42d437, with the reasoning recorded in the header so it isn't rediscovered without data that separates the two cases.

### What ships

Dwell 3 → 2 only. Halves the detector's latency, which the original measurement showed *was* the entire latency, and takes the same second off resume — the "enough to travel across a junction" complaint.

The evidence base went from one tunnel-heavy ride to 32 stops across an urban commute, a 10 km/h pootle and a 53 km/h road ride. Thresholds unchanged and now corroborated by all three riders.

### Corrections to earlier comments in this thread

Two claims I made above were wrong and are superseded:

1. *"The fast path only ever triggers on GPS artefacts."* Wrong — clean rides show the same single-sample collapses; they simply landed on the other side of the ceiling.
2. *"Raising the ceiling to 3.0 is worth it."* That held on five rides and did not survive nine. The extra firings it unlocked were mostly false.

The 30% repeated-value rate quoted early on was also specific to repeated tunnel transits (48% in that section, 12% before it). Normal is 7–13%, so there is no general GPS-hold problem to chase.

PR #292, opened 2026-08-21
fix(cmake): stop claiming a glance interface for every app

sdvsaienko:
## Summary

`una_app_build_app()` passed `-glance_capable` to `app_merging.py` for every
app, so every `.uapp` set the glance-capable bit (bit 5, `0x20`) whether or not
the app serves glance data. None of the non-Glance example apps do — only the
six `Glance*` apps reference `SDK::Glance` at all.

The flag now follows `APP_TYPE`, and is overridable:

| `APP_TYPE`                     | `-glance_capable` |
|--------------------------------|-------------------|
| `Glance`                       | yes (unchanged)   |
| everything else                | no                |
| any, with `APP_GLANCE_INTF` on | yes               |

`APP_GLANCE_INTF` follows the same shape as the existing `APP_AUTOSTART` and
`APP_USE_ICONS` knobs, and is what a non-Glance app that genuinely serves glance
data should set. The second commit brings all three under one reader that
validates what it is given.

## Why it matters

A packed header states two independent things: the app's **type** (`-type`, one
of `Activity` / `Utility` / `Glance` / `Clockface`) and whether the app exposes
a **glance interface** (`-glance_capable`). Emitting the second unconditionally
made every image contradict itself — three of the four types shipped a header
claiming an interface the app has no code for.

The firmware resolves an app's role from those flags, so it has to pick an order
when the two disagree. Under the current order the glance bit outranks the
`Clockface` type: an app packed as `-type Clockface` is taken for a glance and
never routed as a watchface. `Activity` and `Utility` are resolved before the
bit is consulted, which is the only reason this stayed invisible until now.

So a stock SDK cannot currently produce a working watchface, however correctly
the app itself is written.

<sub>Reviewers with firmware access: the precedence sits in the app component's
type-resolution helper.</sub>

## Boolean options are read as booleans, and validated

Second commit. `APP_AUTOSTART`, `APP_USE_ICONS` and `APP_GLANCE_INTF` were read
with `${VAR} STREQUAL On`, which matches the single spelling `On` and reads
every other way of writing the same thing as off:

| written                                 | before           | after           |
|-----------------------------------------|------------------|-----------------|
| `On`                                    | on               | on              |
| `ON`, `on`, `TRUE`, `YES`, `1`          | **off**          | on              |
| `Off`, `OFF`, `off`, `FALSE`, `NO`, `0` | off              | off             |
| `disabled`, `Onn`, `none`               | off              | **build stops** |
| empty                                   | `if()` syntax error | **build stops** |

`-DAPP_GLANCE_INTF=ON` — the spelling CMake's own documentation uses — read as
off, so the override this PR introduces would have quietly done nothing.

`if(VAR)` on its own is not enough. CMake defines a short list of false values
and treats every other value as true, so a typo, or a word like `disabled`,
reads as ON. These options decide what goes into the packed header, so
`una_app_bool_option()` rejects a value that cannot be read as a boolean rather
than picking a side:

```text
APP_USE_ICONS is 'disabled', which is not a boolean; accepted in any case:
ON OFF TRUE FALSE YES NO Y N 1 0
```

The accepted set is narrower than CMake's: an empty value, `NOTFOUND` and
`IGNORE` count as false to CMake, but here they mean the caller wrote the option
and got it wrong. Meson, Bazel and Kconfig reject a non-boolean for a typed
option; CMake accepts anything — `option()` included, which stores
`FOO:BOOL=maybe` and reads it as true — so the check has to be written out.

`${APP_TYPE} STREQUAL Glance` had a second problem: an undefined `APP_TYPE`
expands to nothing, leaving `if()` with two arguments and aborting on `Unknown
arguments specified`. `APP_TYPE STREQUAL "Glance"` evaluates false instead. The
comparison stays case-sensitive, which matches `app_merging.py` — its `-type`
choices are case-sensitive too, so CMake and the packer reject the same input.

## Compatibility

`Glance` apps are deliberately left byte-identical — the claim is true for them,
and they are already in the field.

`Activity` and `Utility` apps lose a bit that was never meaningful for them.
Because their type is resolved before the bit is consulted, nothing about how
they are treated changes; queries that test the bit on its own now answer
correctly for them instead of unconditionally yes.

The second commit changes no packed image at all: every caller in the repository
— `Alarm`, `Timer` and the six `Glance*` apps — already writes exactly `On` or
`Off`.

Worth a look from whoever owns the phone/OTA side, in case anything there keys
off bit `0x20` in the header rather than off the type.

## Verification

Built before and after on this branch:

| App             | `APP_TYPE`  | Before | After  |
|-----------------|-------------|--------|--------|
| `GlanceBattery` | `Glance`    | `0x22` | `0x22` |
| `Stopwatch`     | `Utility`   | `0x21` | `0x01` |
| scratch app     | `Clockface` | `0x23` | `0x03` |

There is no `Clockface` app in `Examples/` yet, so that row comes from a
throwaway app built locally against this branch.

The option handling was driven by extracting `una_app_bool_option()` and the
three option blocks from `una-app.cmake` rather than retyping them, so what ran
is what ships: each of the three options over all twelve accepted spellings, the
glance default still following `APP_TYPE` (`Glance` on, `Activity`, `Utility`
and `Clockface` off), and `disabled`, `Onn` and `none` each stopping the run
before the option is used.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved app packaging so glance capability is automatically configured based on the app type.
  * Non-glance apps no longer receive glance-specific packaging settings by default.
  * Explicit glance interface settings continue to override automatic behavior.
  * Invalid boolean packaging options are now rejected with a clear configuration error instead of being silently accepted.

* **Documentation**
  * Clarified supported app types, boolean values, developer identifiers, and packaging options.
  * Documented glance interface and launcher icon settings, including their defaults.

### DIFF
--- a/Docs/sdk-setup.md
+++ b/Docs/sdk-setup.md
@@ -307,9 +307,19 @@ If build files are generated in the wrong directory (e.g., root instead of build
         - PowerShell/Windows: python -c "import hashlib; print(hashlib.md5(b'MyApp').hexdigest().upper()[:16])"
       - Alternatively, acquire from the developer portal for published apps as described in [deploy.md](deploy.md).
       Replace "MyApp" with your app name. See [API reference](api-reference.rst) for details.
-    - `DEV_ID`: Device identifier, typically "UNAWatch".
-    - `APP_TYPE`: App type, e.g., "app".
-    - `APP_AUTOSTART`: "On" or "Off" (default "Off") to enable autostart.
+    - `DEV_ID`: developer identifier, passed to the app as a compile definition of the same name,
+      which the app may use as it sees fit. Every example and tutorial sets `"UNA"`.
+    - `APP_TYPE`: one of `Activity`, `Utility`, `Glance` or `Clockface`, spelled exactly. It sets
+      the type in the packed header, and the default for `APP_GLANCE_INTF` below.
+    - `APP_AUTOSTART`: enable autostart, default `Off`. Accepted values are `On`, `Off`, `True`,
+      `False`, `Yes`, `No`, `Y`, `N`, `1` and `0`, in any case; anything else stops the build
+      rather than being read as one or the other.
+    - `APP_GLANCE_INTF`: whether the packed header claims a glance interface (bit 5, `0x20`),
+      default following `APP_TYPE`: on for `Glance`, off for every other type. Most apps never
+      set it; set it only for a non-Glance app that genuinely serves glance data. Accepted
+      values as for `APP_AUTOSTART`.
+    - `APP_USE_ICONS`: whether the app ships launcher icons from `RESOURCES_PATH`, default `On`.
+      The `Glance*` examples set it `Off`. Accepted values as for `APP_AUTOSTART`.
     - Paths (adjust if reorganizing file structure):
       - `LIBS_PATH`: Path to libraries, default "../../Libs".
       - `OUTPUT_PATH`: Path for built files, default "../../../Output".
--- a/cmake/una-app.cmake
+++ b/cmake/una-app.cmake
@@ -322,6 +322,33 @@ function(una_app_build_gui TARGET_NAME)
     )
 endfunction()
 
+# Read a boolean app option: apply the default when the caller left it unset,
+# and reject a value that is neither on nor off.
+#
+# CMake defines a short list of false values and treats everything else as
+# true, so an unrecognised value -- a typo, or a word like "disabled" -- reads
+# as ON. These options decide what goes into the packed header, so a value that
+# cannot be read as a boolean stops the build rather than picking a side.
+# The accepted set is narrower than CMake's: an empty value, NOTFOUND and
+# IGNORE all count as false to CMake, but here they mean a caller wrote the
+# option and got it wrong, so they are rejected rather than read as off.
+function(una_app_bool_option NAME DEFAULT)
+    if(NOT DEFINED ${NAME})
+        set(${NAME} ${DEFAULT} PARENT_SCOPE)
+        return()
+    endif()
+
+    # TOUPPER settles the case, so the list is one entry per value, not per
+    # spelling: On, on and ON all arrive here as ON.
+    string(TOUPPER "${${NAME}}" VALUE)
+    set(ACCEPTED ON OFF TRUE FALSE YES NO Y N 1 0)
+    if(NOT VALUE IN_LIST ACCEPTED)
+        list(JOIN ACCEPTED " " SPELLINGS)
+        message(FATAL_ERROR
+            "${NAME} is '${${NAME}}', which is not a boolean; accepted in any case: ${SPELLINGS}")
+    endif()
+endfunction()
+
 # Main function to build a complete watch app
 function(una_app_build_app)
     set(OUTPUT_COPY_COMMANDS "")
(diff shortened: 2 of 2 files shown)


rryles (commented):
Good find, and the write-up is unusually clear about *why* it stayed invisible. I checked the
firmware side against your `<sub>` note and **your reading of the precedence is exactly
right**: the glance bit is consulted ahead of the `Clockface` type, so a `Clockface`-typed
image that also carries the bit never resolves as a watchface, while `Activity` and `Utility`
resolve earlier and are genuinely unaffected. The compatibility argument holds as written.

I also confirmed the second commit is byte-neutral: every caller in the repo (`Alarm`, `Timer`
and the six `Glance*` apps) writes exactly `On` or `Off`, so nothing repacks. And the
validated spellings all resolve correctly under CMake's own truthiness rules — `N`, `NO`, `0`,
`OFF`, `FALSE` are all false constants and case-insensitive, so `if(APP_GLANCE_INTF)` agrees
with `una_app_bool_option()` for every accepted value. `list(JOIN)` and `IN_LIST` are fine
against the 3.21 minimum.

## This gates the clockface work — worth merging first

Since a stock SDK can't currently pack a watchface the firmware will route as one, this is a
prerequisite for #297 and for the kernel-side clockface PR, not just a correctness tidy-up.
Suggest landing it ahead of both so the Analogue face is exercised against correct header
bytes rather than needing a follow-up repack.

There's also a related firmware-side inconsistency in how the two type-resolution helpers
disagree for `Clockface`. It's out of scope here and not something this PR needs to handle —
I'll raise it in the kernel repo separately.

## `APP_GLANCE_INTF` should be documented

`Docs/sdk-setup.md` lists the app options around line 311 — `APP_TYPE`, `APP_AUTOSTART`,
`APP_USER_NAME`, `APP_FILE_NAME`. This adds a new public knob that a third-party app author is
now expected to reach for, so it belongs in that list. Worth noting the default follows
`APP_TYPE`, since that's the surprising part.

While you're in there, the `APP_TYPE` line itself is wrong today:

> `APP_TYPE`: App type, e.g., "app".

`"app"` isn't one of the four accepted values. That was merely unhelpful before; now that
`APP_TYPE` silently decides the glance default, a reader following it lands on the wrong
behaviour with no error. Good moment to fix it to `Activity` / `Utility` / `Glance` /
`Clockface`.

(`APP_AUTOSTART` is also documented as `"On" or "Off"`, which this PR quietly widens — worth a
few words there too, though that one's cosmetic.)

## Optional: give `APP_TYPE` the same treatment

The booleans now fail loudly on a bad value, but `APP_TYPE` — which the glance default hinges
on via an exact, case-sensitive match — still doesn't. A typo like `clockface` silently takes
the non-Glance branch and then dies further down in `app_merging.py`, away from the line that
caused it. Validating it against the same four values the packer accepts would keep the
failure local and legible. Entirely your call; the packer does eventually catch it.

## On the open question

Kernel side I can confirm nothing keys off bit `0x20` in a way this changes — the only
consumers are the type resolution itself and the alias rules, which force-clear the bit
anyway. The phone/OTA side still needs someone with that repo to confirm before this lands.

No objection to the change itself — the docs additions are the only thing I'd want before
merge, and I'm happy for you to fold those in and land it without another round trip from me
once the phone/OTA side is confirmed.

rryles (approved):
Docs commit looks good, and you went further than I asked — `DEV_ID` and `APP_USE_ICONS` were
both wrong or missing and I hadn't spotted either. I checked the new entries against the code
rather than the prose: the four type names match the packer's `APP_TYPES` keys exactly, the
ten boolean spellings match `una_app_bool_option()`, and all 23 files setting `DEV_ID` do use
`"UNA"`.

The phone/OTA question is answered too — confirmed with the phone side that nothing there
reads bit `0x20`; their glance classification doesn't come from the header at all. Your
follow-up about a mixed fleet during rollout is moot for the same reason. Thanks for chasing
that rather than leaving it on the PR.

That clears everything I had. Approving.

Two things deliberately not blocking:

- The optional `APP_TYPE` validation is still your call — the packer catches a bad value
  eventually, it just does it further from the cause.
- Needs a rebase now that #294 has landed.

PR #293, opened 2026-08-25
fix(scrollindicator): keep both wrap handles on the rail

sdvsaienko:
During a wrap-around slide both handles are placed from raw angles that fall past
the ends of the rail. The container hides most of that overshoot, but not all of
it: a narrow band beyond each rail end still renders, so the handle is drawn
starting just outside the rail rather than at its end. This fixes that, in the
widget template and in the app copies still carrying the old code.

## Problem

`ScrollIndicator` ships as a template and every app keeps its own copy, free to
adapt it. These six had not adapted it -- they were carrying the template's code
verbatim, bug included, and so was the template itself:

| copy | carries the fix |
|---|---|
| `Timer` | yes |
| `Cycling`, `Hiking`, `Running`, `Stopwatch`, `Treadmill`, `Workout` | no |
| `Docs/Templates/TouchGFX-Widgets/ScrollIndicator.tpkg` | **no** |

The template is the part that keeps costing: an app starts from whatever it holds,
which is how six copies came to carry this one.

`getStartAngle()` runs high-to-low -- index 0 sits at `railMax - handleLen`. On a
backward wrap (`animateToId(-1)`, kBig, 5 items) the incoming handle is armed at
`232 - 18 = 214`, below `railMin`, and the outgoing one runs to `290 + 18 = 308`
with its far edge at 326, above `railMax`. Both were placed there unclipped.

The container hides most of that. With the Designer geometry -- `setCenter(120, 73)`,
`setRadius(112)`, `setLineWidth(8.5)` inside a 36x146 box -- only `[229.4, 310.6]`
reaches the screen at all. But the rail is `[232, 308]`, so a 2.6 degree band past
each end does render: roughly 5 px of arc at an 8.5 px stroke, drawn beyond the
grey rail. The handle is never missing, and it grows monotonically throughout the
slide; it simply starts a sliver early and ends a sliver late.

## Approach

`setClampedArc()` clips both handles to `[railMin, railMax]` and hides an arc that
clamps to nothing, so the outgoing handle shrinks into one rail end while the
incoming grows out of the other. For the normal in-rail positions it is a no-op.
Both bounds are clamped both ways, so the invariant holds in code rather than by
argument about which call sites exist.

The clip is presentation only, and that distinction is load-bearing. The widget
used to read its own position back out of the rendered arc (`getArcStart()`), so
clipping the render would have silently changed the interrupt path: interrupting a
backward wrap mid-slide would resume from the rail edge instead of from where the
handle logically was. The unclipped position now lives in `mHandlePos` / `mOvfPos`
and the render is derived from it. This also drops a round-trip through `Circle`'s
Q5 fixed-point quantisation.

`Timer` already had the clip and is not part of the back-port, but the position
rework lands there too -- which is why the diff touches seven apps rather than the
six listed above.

`ScrollIndicatorBase.*` was already byte-identical across every copy -- the
Designer definition never drifted, only hand-written code did. The repacked
`.tpkg` therefore differs in its two source entries plus the `<CreatedAt>` stamp
in `manifest.xml`; the container export, `CustomContainerManifest.xml`,
`texts.xml` and the entry order are untouched.

`docs/ScrollIndicator.md` gains two notes. The public API is unchanged;
`setClampedArc()` is private.

## Tested

- `Cycling` simulator and the `Cycling` `.uapp` on hardware -- wrap-around in the
  main menu, both directions. No regression.
- `Stopwatch` simulator -- it clamps its own scroll and never reaches the wrap
  path, so its behaviour must be unchanged, and is.
- Simulators build clean, no warnings.

Note on coverage: what the fix removes is the 2.6 degree sliver described above --
about 5 px of arc at each rail end. An A/B capture of the simulator was tried and
did not isolate it: two runs cannot be held at the same point in the animation, so
a difference in phase is indistinguishable from a difference in behaviour at that
scale. The runs above establish that nothing regressed; the case for the change
itself is the geometry, not a screenshot.

The `mHandlePos` / `mOvfPos` rework is narrower still and is not separately
observable. It exists because clipping the render would otherwise have changed the
interrupt path: the widget read its own position back out of the rendered arc, and
a clipped arc reports the rail edge instead of where the handle is.

## Follow-up

A copy has no way to tell whether it predates a fix in the template it came from.
`manifest.xml` records a `MinimumTouchGFXVersion` and nothing about the widget
itself, a `.tpkg` is a binary, and the sources do not say which revision they are.
Adapting a copy is expected and fine -- but a developer who has adapted one still
wants to know when a fix lands upstream, and today that means diffing by hand
against the SDK.

A revision line in the widget sources would turn that into comparing two values,
the copy's and the template's:

```cpp
// SDK widget template ScrollIndicator, revision 2026-08-25.
// Local changes to this file are expected; a newer revision may exist upstream.
```

The line has to read correctly from both ends, because the same bytes are both
things: there is no plaintext master, the template's source exists only inside the
`.tpkg`. Naming the template and its revision does that -- in the template it
states its own revision, in a copy it says which revision the file came from -- and
neither reading stops being true once the copy is adapted. The template side of the
comparison already exists, too: `manifest.xml` carries a `<CreatedAt>` that the
Designer restamps on every export, so the line can be taken from it rather than
maintained separately.

It travels inside the `.tpkg` for free, because the Designer carries the source
files verbatim. What is worth settling first is the granularity that keeps the two
in step -- a date works, since the sources are written before the export that
stamps them -- and what the docs record alongside it. This applies to every widget
template rather than to this one, so: its own PR.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved scroll indicator wrap-around animations so handles remain clipped within the rail.
  - Interrupted animations now resume from the actual handle position instead of the clipped visual position.
  - Corrected initial animation behavior when scrolling begins before item counts are set.
  - Improved visibility handling for zero-length or fully clipped handle arcs.

- **Documentation**
  - Clarified scroll indicator animation and clipping behavior.


rryles (dismissed):
Good fix, and the analysis of *why* it was visible at all — the container hides most of the
overshoot but leaves a 2.6 degree band, roughly 5 px at an 8.5 px stroke — is the part that makes
this reviewable rather than a plausible-looking clamp.

The genuinely important half is separating the logical position from the rendered one.
Clipping the render while the widget still read its position back out of `getArcStart()` would
have quietly changed the interrupt path, and that is a much worse bug than the sliver being
fixed. Moving the truth into `mHandlePos` / `mOvfPos` and deriving the render from it is right,
and dropping the round-trip through `Circle`'s Q5 quantisation is a free bonus.

`setClampedArc()` clamping both bounds both ways, rather than relying on which call sites exist
today, is the right instinct for a widget that gets copied into apps and edited.

## Verified

- **All seven app copies are now byte-identical** — I md5'd every one at the head, `.cpp` and
  `.hpp`. They also match **the kernel's copy** on `feature/clockface-support` exactly, so all
  eight instances of this widget now agree, with the `Timer` divergence fully converged.
- **The `.tpkg` really was repacked with the fix.** This is the claim that matters most, since
  the PR's thesis is that the template is what keeps costing — a repack that silently missed
  would let the bug back in through the next app. I extracted it and compared: the two source
  entries differ from the app copies only by CRLF line endings (7617 vs 7368 bytes, exactly 249
  extra bytes for a 249-line file), and are otherwise identical.
- At the merge base, six copies were byte-identical and `Timer` differed only by the clip, which
  matches the table in the description.

## One thing to fix while you are here

**`*.tpkg` has no `binary` rule in `.gitattributes`.** There are twelve of them under
`Docs/Templates/TouchGFX-Widgets`, and they currently fall under the repo-wide `* text=auto`,
relying on git's NUL-byte sniffing to guess right. It does guess right — a zip is unambiguous —
but that is luck rather than intent, and it is exactly the reasoning you used in #298 to add
`*.fig binary`. Same one-line rule, and this PR is the one modifying a `.tpkg`:

```
*.tpkg   binary
```

Approving. Worth a rebase past #294 before it lands.

sdvsaienko:
Good catch, and the `*.fig` precedent is the right frame -- the rule that keeps
these intact should be stated, not inferred from a NUL byte. Added:

    *.tpkg   binary

Confirmed it takes effect (`git check-attr` now reports `binary: set` instead of
`text: auto`) and that adding the rule does not renormalise the twelve existing
packages -- only `.gitattributes` shows as modified.

PR #294, opened 2026-08-25
fix(gui): put SensorStatusRow in SDK::GUI with the rest

sdvsaienko:
`SensorStatusRow` sits in `SDK::Gui`, while everything else under
`Libs/Header/SDK/GUI/` is in `SDK::GUI`. Two namespaces that differ only in case,
in one directory, is a trap waiting for the next person to add a file there. This
moves the odd one out.

## Why this direction

| | declared in | uses | files |
|---|---|---|---|
| `SDK::GUI` | `Button.hpp`, `Color.hpp`, `Config.hpp` | 468 | 131 |
| `SDK::Gui` | `SensorStatusRow.hpp` | 53 | 30 |

`SDK::GUI` also came first -- the three headers landed together in `fe911165`
(2026-04-26), `SensorStatusRow` two months later in `9a43b240`, with both a
different case and the older nested-namespace syntax. It reads as a slip rather
than a second convention: nothing else was ever added to `SDK::Gui`, and the SDK
library itself never references it (`TouchGFXCommandProcessor.cpp` and the
simulator `Kernel.cpp` both reach for `SDK::GUI::Button`).

The declaration adopts the C++17 form its siblings use. The class stays one level
up, because `SDK::GUI::Button` and friends are namespaces of constants while this
is a type:

```cpp
namespace SDK::GUI
{
class SensorStatusRow : public touchgfx::Container { ... };
} // namespace SDK::GUI
```

## The header guard, while the file is open

`__SDK_GUI_SENSOR_STATUS_ROW_HPP` also loses its leading underscores. Identifiers
beginning with a double underscore are reserved to the implementation, so that
name was never ours to define. The macro appears only in this file, so nothing
else moves.

## Scope

31 files: the header plus 30 call sites across `Cycling`, `Hiking`, `Running`,
`Treadmill` and `Workout`. Nothing else in the repository refers to the name --
not the tutorials, not the docs, and no other example app.

Only the fully qualified `SDK::Gui::` was replaced, never a bare `Gui`. Every app
has its own unrelated `Gui` namespace in `gui/common/GuiConfig.hpp`, which the
tutorials document as `Gui::Config::...`; a careless rename would have taken that
with it.

## Tested

- `Cycling` and `Workout` simulators build clean. Both use the widget.
- `git grep SDK::Gui` returns nothing.
- Behaviour is untouched: the diff changes namespace and guard lines only.

### DIFF
--- a/Examples/Apps/Cycling/Software/Apps/TouchGFX-GUI/gui/include/gui/containers/MainMenuLayout.hpp
+++ b/Examples/Apps/Cycling/Software/Apps/TouchGFX-GUI/gui/include/gui/containers/MainMenuLayout.hpp
@@ -84,7 +84,7 @@ public:
     MainMenu &getMenu();
 
 private:
-    SDK::Gui::SensorStatusRow mSensorRow;
+    SDK::GUI::SensorStatusRow mSensorRow;
 };
 
 #endif // MAINMENULAYOUT_HPP
--- a/Examples/Apps/Cycling/Software/Apps/TouchGFX-GUI/gui/include/gui/containers/TrackFaceStatus.hpp
+++ b/Examples/Apps/Cycling/Software/Apps/TouchGFX-GUI/gui/include/gui/containers/TrackFaceStatus.hpp
@@ -33,11 +33,11 @@ public:
     void setBatteryLevel(uint8_t level);
 
     /** @brief Set the GPS / external-HR icon states (see SensorStatusRow). */
-    void setGps(SDK::Gui::SensorStatusRow::State state) { mSensorRow.setGps(state); }
-    void setHr(SDK::Gui::SensorStatusRow::State state)  { mSensorRow.setHr(state); }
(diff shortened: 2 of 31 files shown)


rryles (approved):
Clean move, and the reasoning for picking `SDK::GUI` over `SDK::Gui` is the right way round —
usage, precedence and the fact that the SDK library itself only ever reaches for `SDK::GUI`.

Checked independently:

- **No cross-repo breakage.** `SDK::Gui` and `SensorStatusRow` appear nowhere in the kernel
  repo, so nothing breaks when the submodule pointer moves past this. That was the one risk
  worth ruling out, since "nothing else in the repository" naturally means this repo.
- Guard rename matches the convention — no leading underscores, and the `#endif` comment
  tracks the new name.
- Only the qualified `SDK::Gui::` is touched; each app's unrelated `Gui` namespace is intact.

Nice catch on two namespaces differing only in case sitting in one directory. Approving.

PR #296, opened 2026-08-25
refactor(gui): share one CountdownTimer instead of seven copies

sdvsaienko:
`CountdownTimer` was copy-pasted into every app. The copies were identical apart
from one defect, so this replaces them with a single header-only widget in
`SDK::GUI`, next to `SensorStatusRow`.

## Why this widget and not the others

Widget templates are meant to be copied and adapted -- `Buttons` legitimately
differs per app (`TEAL` for utilities, `GREEN` for activities), `MainMenu` carries
a per-app special case. That freedom is the point, and consolidating those would
fight it.

`CountdownTimer` is not one of them. It was never distributed at all: there is no
`CountdownTimer.tpkg` in `Docs/Templates/TouchGFX-Widgets` and no entry in
`touchgfx-widgets.rst`. It was copy-pasted. And there is nothing in it to adapt --
no Designer base, no geometry, no colours; the whole surface is `setDuration` /
`setCallback` / `start` / `stop`. All seven copies were byte-identical except for
the defect below.

## The defect

**A zero-duration countdown never expired.** The tick guard was `if (mCounter ==
0) return;`, so a timer started with `mDuration == 0` stayed armed forever and the
callback never fired. Decrementing from `0` would also have underflowed the
`uint16`. `Timer` already carried the fix; the other six did not.

Nothing reaches it today: all 36 `setDuration` call sites pass a fixed
`secToTicks(1..120, kFrameRate)` constant. This closes a latent hole rather than
a live bug, which is also why the change is invisible at every call site.

## Naming the running state

`mRegistered` tracked registration with the framework, and nothing exposed whether
a countdown was in progress. The shared widget adds `isRunning()`, and that needs
the state to be explicit rather than inferred -- deriving it from the counter is
wrong in two of the four reachable states:

| running | `mCounter` | when | `mCounter > 0` would say |
|---|---|---|---|
| yes | > 0 | ordinary countdown | yes |
| **yes** | 0 | zero-duration, before its first tick | no |
| **no** | > 0 | after `stop()`, which leaves the counter | yes |
| no | 0 | after firing | no |

So `mRegistered` becomes `mRunning`, and one place changes it -- the same place
that registers and unregisters, so the flag and the framework cannot drift apart:

```cpp
void setRunning(bool running)
{
    if (running == mRunning) {
        return;
    }

    mRunning = running;
    if (running) {
        touchgfx::Application::getInstance()->registerTimerWidget(this);
    } else {
        touchgfx::Application::getInstance()->unregisterTimerWidget(this);
    }
}
```

The tick guard keys off the same flag, which is what lets `isRunning()` simply
return it.

`isRunning()` has no caller in this repository. It is here because the running
state is now tracked explicitly either way, and a widget that can be started and
stopped should be able to answer whether it is running.

## reset()

Unchanged, and still not called anywhere -- but kept. It pushes the deadline back
without touching the running state, which is a real pattern (extend an
auto-dismiss while the user is active). Naming the state turned its precondition
into something the code says by itself: a stopped timer stays stopped, and
`start()` arms one again.

## Copying

Deleted, which the per-app copies never did. The framework holds `this` while the
timer runs, so a copy would carry `mRunning` without ever having registered, and
its destructor would unregister a pointer the framework never had. `Doxyfile` has
`INPUT = ../Libs/Header` and `RECURSIVE = YES`, so this header lands in the
published API reference -- the constraint belongs in it.

## Scope

68 files:

| | |
|---|---|
| new shared header | 1 |
| deleted copies | 14 |
| consumer headers | 34 |
| simulator project files | 14 |
| architecture docs | 5 |

The header is self-contained: it includes `touchgfx/Application.hpp` itself, which
every consumer already had, so no translation unit gains anything.

The firmware build picks up `gui/src/*.cpp` by glob, so no firmware source list
needed editing; the simulator projects listed the `.cpp` explicitly and were
updated. An existing local build tree may have to be re-configured after the
deletion, as it would after deleting any other file.

## Tested

- `Alarm`, `Timer` and `Workout` simulators build clean.
- `Alarm` `.uapp` builds, confirming the firmware glob picked up the deletion.
- No unit test: `Tests/Host` does not link TouchGFX, so this widget cannot be
  exercised there without new infrastructure. The two states the change touches
  are covered by reasoning and by the call-site survey above, not by a test.

## Summary by CodeRabbit

* **New Features**
  * Added a shared countdown timer widget with duration, start, stop, reset, running-state, and callback support.
  * Countdown timers now stop automatically before triggering callbacks when time expires.

* **Refactor**
  * Updated Alarm, Cycling, Hiking, Running, Timer, Treadmill, and Workout interfaces to use the shared timer.

* **Documentation**
  * Updated architecture examples to identify the shared countdown timer and its source.


rryles (dismissed):
Agreed on the criterion, which is the part worth getting right: `Buttons` and `MainMenu` are
meant to differ per app, and consolidating those would fight the design. `CountdownTimer` is a
different animal — never distributed as a template, no Designer base, no geometry or colours,
nothing to adapt. Copy-paste rather than deliberate divergence, and worth collapsing.

## Verified the consolidation loses nothing

This is the risk with a change like this — an app that had quietly adapted its copy gets its
adaptation deleted — so I checked rather than trusted. At the merge base:

- `Alarm`, `Cycling`, `Hiking`, `Running`, `Treadmill`, `Workout` — all six byte-identical.
- `Timer` — differs by exactly the guard, nothing else. I diffed it: the whole delta is
  `if (mCounter == 0)` → `if (!mRegistered)` plus the `mCounter > 0 && --mCounter != 0` rework.
- `Stopwatch` — no copy at all.

So seven copies, no adaptations, and the shared widget is a superset of every one of them.

## The tick logic is right

```cpp
if (mCounter > 0 && --mCounter != 0) { return; }
```

Zero-duration falls straight through and fires on the first tick; `1` decrements to zero and
fires; `N` fires after N ticks. No underflow path.

`setRunning(false)` running *before* `mpCb->execute()` is the detail I would have looked for:
it means `isRunning()` reads false inside the callback, and a callback that calls `start()`
re-arms cleanly instead of being unregistered a moment later. Worth keeping that ordering
comment if anyone ever reshuffles the function.

The state table in the description earns its place — deriving `isRunning()` from `mCounter > 0`
really is wrong in two of the four reachable states, and it is not obvious until written out.
Funnelling both the flag and the registration through `setRunning()` is the right shape.

Deleting the copy constructor and assignment, with the reason stated, is a real improvement on
what the copies had: the framework holds `this`, so a copy carrying `mRunning` without ever
having been registered would have unregistered a pointer the framework never had. That was
latent in all seven.

## Two small ones, neither blocking

- **Say why it is a `Container`.** It has no children, no geometry and draws nothing; it derives
  from `Container` purely so `registerTimerWidget()` will take it. That is obvious once you know
  and puzzling until you do — one line in the class comment would save the next reader the
  detour.
- **Does a shared `SDK::GUI` widget need a documented home?** The argument for consolidating is
  partly that this was never in `touchgfx-widgets.rst` or shipped as a `.tpkg`. Now that it is a
  real shared widget alongside `SensorStatusRow`, there is a category — SDK-provided widgets you
  include rather than copy — that the docs do not describe anywhere. Not this PR's job to create
  it, but worth deciding before a third one lands.

Approving.

sdvsaienko:
Thanks for diffing the copies at the merge base -- an app that had quietly adapted
its copy was the one way this could have been lossy. Both points are in the push.

**Why it is a Container.** The explanation was in the file banner, which is the
wrong place twice over. `Doxyfile` has `INPUT = ../Libs/Header`, and a file-level
`@brief` renders on the file page, not the class page, so in the API reference the
class carried no explanation at all. And "it derives from Container so
`registerTimerWidget()` will take it" answers the wrong question, since that takes
a `Drawable`. The class comment now says why not `Drawable`: it is abstract,
`Container` adds one member over it (`firstChild`) and implements `draw()`,
`getSolidRect()` and `getLastChild()` sensibly for something with no children -- a
bespoke subclass would save four bytes at the cost of three hand-written
overrides.

**Ordering.** Commented, as you suggested: the countdown leaves the running state
before the callback runs, so `isRunning()` is false inside it and a callback that
calls `start()` re-arms instead of being unregistered a moment later.

**A home for shared widgets.** Sharper than the docs not describing it --
`touchgfx-widgets.rst` opens with "Each widget is distributed as a `.tpkg` file",
so header-only widgets are excluded by its premise rather than merely absent.
`SensorStatusRow` has been in that category since June and is not mentioned
either. Two members, no home. Raising it separately.

**Before merge.** Rebasing onto `main` drops the duplicate `SensorStatusRow`
commit rather than conflicting: it and the merged `4b98299e` have the same
patch-id.

rryles (approved):
Both addressed, and the first one corrects me rather than just doing as asked — which is the
better outcome.

**The `Container` explanation.** You are right that "so `registerTimerWidget()` will take it" is
the wrong answer, and I checked the framework rather than take the correction on trust:
`registerTimerWidget(Drawable* w)` does take a `Drawable`; `Drawable` is abstract on exactly the
three you name — `draw()`, `getSolidRect()`, `getLastChild()`; and `Container` adds exactly one
member, `Drawable* firstChild`, while implementing all three. So the comment as written is
accurate, and it answers the question a reader actually has. The point about the file banner
missing the class page under `INPUT = ../Libs/Header` is a good catch on top — that would have
left the class undocumented in the generated reference no matter how good the prose was.

"Not worth four bytes" is the right call, and worth having written down so nobody relitigates it.

**The ordering comment** says what it needs to.

**Rebase.** Confirmed from here: the branch is now a single commit and `SensorStatusRow.hpp` has
dropped out of the diff entirely, so the duplicate really did fold into the merged `4b98299e`
rather than conflict. Only `Libs/Header/SDK/GUI/CountdownTimer.hpp` remains outside the app
trees. `MERGEABLE/CLEAN`, checks green.

**On the missing home for shared widgets** — agreed, and your framing is sharper than mine.
`touchgfx-widgets.rst` does open with "Each widget is distributed as a `.tpkg` file", so a
header-only widget is excluded by the page's premise, not merely absent from its list. I
confirmed `SensorStatusRow` appears nowhere in it. Two members and no home is the right moment
to give the category one, and a separate issue is the right place. Worth saying there that the
two kinds differ in more than packaging: a `.tpkg` is copied and adapted, an `SDK::GUI` header is
included and shared, and that distinction is what decides whether a fix has to be back-ported to
seven apps or lands once — which is the whole argument of this PR.

Nothing else changed; the logic is as reviewed. Approving.

PR #298, opened 2026-08-29
docs(templates): add the Figma UI resource pack

sdvsaienko:
## What

Adds the UNA Watch UI Resource Pack (Figma, v1.0, exported August 2026) to the
SDK and documents it:

- `Docs/Templates/Figma-UI-Kit/UNA-Watch-UI-Resource-Pack.fig` - the pack itself
  (5.1 MB), plus a short `README.md` for anyone browsing the repo.
- `Docs/figma-ui-kit.rst` - a docs-site page: what is in the pack, how to import
  it into Figma, and the rules for getting a Figma screen into a TouchGFX app.
  Added to `Docs/index.rst` under Architecture, next to `touchgfx-widgets`.
- `.gitattributes` - `*.fig binary`, so the repo-wide `text=auto` rule never
  rewrites the archive.

## Why

The SDK already ships code-side templates (`Docs/Templates/TouchGFX-Widgets`),
but nothing on the design side. An app author starting a screen had no watch
frame, no button labels in place, and no reference for the visual language of
the built-in apps, so third-party apps drift away from the system look for no
good reason. This closes that gap with the same source the built-in GUI uses.

## Notes

- The `.fig` is a 5.1 MB binary and this repo has no Git LFS. Every future pack
  revision adds a full copy to history. Acceptable for a file that changes a few
  times a year; worth revisiting if the pack starts moving often.
- The docs page links the file with Sphinx `:download:`, so the pack is also
  downloadable from the published site.
- Docs-only change: no source, no build input. `docs.yml` runs on `Docs/**` and
  will build the new page in this PR; not built locally (no Sphinx toolchain on
  the machine).

## Summary by CodeRabbit

- **Documentation**
  - Added guidance for the UNA Watch Figma UI Resource Pack, including its contents, download and import instructions, and design-to-TouchGFX workflow.
  - Documented the pack as a design reference for watch mockups, branding, screens, and built-in app UI elements.
  - Added the new Figma resource documentation to the main documentation index.
- **Chores**
  - Updated asset handling so Figma files are treated as binary resources.


rryles:
Docs page and `.gitattributes` handling both look right, and I appreciate you flagging the
repo-weight cost up front rather than letting it land quietly. One option worth considering
before this merges, since it's much cheaper to change now than later.

## Attach the pack to a Release instead of committing it

Upload `UNA-Watch-UI-Resource-Pack.fig` as an asset on an `una-sdk` release and have
`figma-ui-kit.rst` link that URL, rather than `:download:`-ing a tracked file.

- Zero clone weight, and future revisions don't accumulate in history at all — each one is
  just a new asset.
- No `git-lfs` dependency for external app developers, who are the main audience for this
  repo.
- The docs page reads almost the same; only the `:download:` role becomes an external link.

The trade-off is that the pack is no longer versioned alongside the code it matches. For a
design reference that no build reads, that seems like a fair price — but it's a real
downside and your call, especially if you expect the pack to track GUI changes closely.

## Why I'd steer away from Git LFS specifically

LFS is the obvious alternative, but it doesn't fit this repo as well as it looks:

- **No checkout step in either repo sets `lfs: true`** — 9 in una-sdk, 3 in una-kernel. With
  LFS enabled, `docs.yml` would copy the ~130-byte pointer file into `_downloads` and publish
  *that* as the `.fig`. The build passes and nothing warns; the download is just silently
  corrupt. It's fixable (`lfs: true` on the docs build checkout), but it's an easy trap to
  leave behind.
- Every SDK consumer would need `git-lfs` installed or they get a pointer file instead of the
  pack — quiet and confusing rather than a clear error.
- `actions/checkout`'s `lfs: true` doesn't recurse into submodules, so the kernel side would
  need an explicit `git lfs pull --recurse-submodules` if it ever wanted the file, and
  `worktree-setup.ps1` only does `submodule update --init`.

## If it does stay in-tree

Then what's here is fine as-is and the note in the description is the right way to handle it.

Worth knowing though: **now is the only cheap moment to switch to LFS.** While the `.fig` is
unmerged it's a one-line `.gitattributes` change. Once it's on `main` as a normal blob, the
only fix is `git lfs migrate import`, which rewrites history — a non-starter with FF-only
`main` and the number of live worktrees. So it's worth deciding deliberately now rather than
defaulting into it.

No objection to merging as written if you'd rather keep the pack versioned with the code.

sdvsaienko:
Thanks - the LFS analysis is the useful part here, and I agree with the conclusion.

**No LFS.** Your pointer-file trap is the deciding argument: I checked, and no checkout
step in either repo sets `lfs: true`, so the docs build would copy a ~130-byte pointer into
`_downloads` and publish it under the `.fig` name. A green build serving a corrupt download
is exactly the failure mode we don't want, and the fix would have to be remembered in every
future workflow that touches these files.

**Keeping the pack in-tree**, deliberately rather than by default, since you're right that
this is the cheap moment to decide.

What steered me away from the release-asset option is that nothing would produce the asset.
Release assets here are built by CI from repo content - `tutorials-ci.yml` zips `artifacts/`
and hands the zip to `action-gh-release`. A design file isn't built from anything, so it
would need a human upload per release, and a step that only works when someone remembers it
doesn't work. It also means the docs link points at the pack from the last release rather
than the pack the current docs describe.

I also looked at hosting it on our own portal and linking that URL. It's the most promising
alternative, but not yet: it needs a scriptable, unauthenticated upload path, and the docs
build runs plain `sphinx-build` with no `linkcheck`, so an external URL that moves would 404
silently for as long as nobody notices. A tracked file can't reach that state.

On the weight itself: 5.1 MB against a repo that's already ~101 MB, for a file that changes
a few times a year. The machinery to avoid that costs more than the 5%.

The escape hatch stays open precisely because it isn't LFS - if a real design-asset section
appears (more packs, icons, fonts), moving this out is a plain file move plus a link change,
with no history rewrite.

rryles (approved):
Convinced — and by the two arguments I hadn't weighed properly.

**Nothing would produce the release asset.** I checked and you're right about the shape of it:
`tutorials-ci.yml` downloads build artifacts, zips them, and hands the zip to
`action-gh-release`. Every asset in this repo comes out of CI from repo content. A design file
has no producer, so it would need a human upload per release — and you're right that a step
which only works when someone remembers it doesn't work. The related point is the one I'd
missed entirely: the docs would then link the pack from the *last* release rather than the one
the current page describes, which is a correctness problem, not just an inconvenience.

**The portal option has my own failure mode.** Also confirmed: the build is a plain
`sphinx-build Docs _build/html` with no `linkcheck`, so an external URL that moves 404s
silently for as long as nobody notices. That's the same class of failure I warned about with
the LFS pointer file, which makes it a fair counter — I was weighing that risk on one option
and not the other.

So: in-tree, deliberately, and the reasoning is now on the record for whoever revisits it. The
escape hatch staying cheap *because* it isn't LFS is the right way round — a plain file move
and a link change beats a history rewrite, and that's worth more than the 5%.

Approving. Worth a rebase past #294 before it lands.

PR #304, opened 2026-09-04
feat(sensors): add MAGNETIC_FIELD and MAGNETIC_FIELD_RAW parsers

sdvsaienko:
### Why this comes first

`MAGNETIC_FIELD` has been in `SDK::Sensor::Type` since the enum was written,
but nothing has ever produced it and it has no parser: the docs list it as
`Parser Available: No`, and the one place that reads a magnetic sample -- the
Sensors tutorial -- pulls floats out of a `DataView` by index and works the
angle out by hand.

The kernel side is what changes that. `una-kernel` gains a sensor-layer
binding that publishes both streams, and consumes these parsers directly.
**That work cannot land until this does** -- it includes
`SensorDataParserMagneticField.hpp` and names `MAGNETIC_FIELD_RAW`, neither of
which exists on `main` today. The kernel-repo PR follows, with the submodule
pointer bumped onto this.

This merges on its own: it depends on no kernel change and waits for nothing.
The stream it parses arrives with the kernel PR.

### What is in it

**A new type.** `MAGNETIC_FIELD_RAW = 0x31`, sitting next to `MAGNETIC_FIELD`
the way `ACCELEROMETER_RAW` and `GYROSCOPE_RAW` sit next to theirs. Purely
additive: no existing value moves.

**Two parsers.**

| Type | Fields |
|------|--------|
| `MAGNETIC_FIELD` (0x30) | `MAG_X`, `MAG_Y`, `MAG_Z` (float uT), `MAG_CALIBRATED` (u32) |
| `MAGNETIC_FIELD_RAW` (0x31) | `MAG_X`, `MAG_Y`, `MAG_Z` (float uT) |

The corrected stream carries a calibrated flag because the distinction is not
cosmetic. A watch that has never been calibrated still produces samples: they
carry the part's own hard-iron offsets, which can be larger than the Earth's
field itself. That is exactly what a calibration is derived from, and exactly
what a bearing must not be taken from. Without the flag a consumer cannot tell
the two apart, and the failure is silent -- a compass that points confidently
in the wrong direction.

`MAGNETIC_FIELD_RAW` deliberately offers no bearing at all, for the same
reason.

**A compass bearing, computed in the parser rather than published as a
stream.** A level heading is a function of one sample and nothing else -- no
state, no history -- so a separate sensor type would have meant another
driver, another subscription and another queue write per sample to carry an
`atan2`. `getAzimuthDeg()` gives degrees clockwise from magnetic north, and
means nothing unless `isAzimuthValid()`.

A tilted one needs gravity as well, and that is exactly why it is an argument
rather than a subscription: the parser gains a second input and still no
state, no history and no stream of its own. Where the caller reads the
accelerometer is the caller's business.

```cpp
float bearing = 0.0f;
if (p.getAzimuthDegTilted(ax, ay, az, bearing)) {
    // read the accelerometer wherever you already read it
}
```

Both routes share one last step, so a watch held flat reads the same either
way -- the level case is the tilted case with zero tilt, not a second
implementation of it.

**Docs.** `Docs/SensorsLayer.md` gains rows for both types and a full section
for each, in the shape the other types use. The "for sensors without parsers
(e.g. `MAGNETIC_FIELD`)" example now names `HEART_BEAT` instead, since the
original example became false with this change -- and it was the passage
anyone wanting a compass would have read.

**Tutorial.** `Docs/Tutorials/Sensors` computed `atan2f(y, x)`: the
mathematical angle anticlockwise from +X, where the parser gives a bearing
clockwise from +Y. Ninety degrees out and turning the wrong way -- and taken
from a field that may be uncalibrated, which the parser's own contract says
must not be done. It now uses the parser and checks validity. `ARCHITECTURE.md`
follows.

### Testing

16 new tests across the two parsers, and the full host suite is green. Built
with `-Wall -Wextra -Wpedantic`, no warnings. `check_msvs_sources.py` clean.

Two of the tests are worth calling out, because the tilt maths is the part of
this that can be quietly wrong:

- **`ProjectingGravityItselfLeavesNothingHorizontal`** feeds gravity in as if
  it were the field, over a grid of attitudes. Gravity points straight down by
  definition, so a transform that puts the horizontal plane where the
  horizontal plane really is must leave it no horizontal component. This
  depends on no Euler convention and catches any disagreement between the
  angles extracted from gravity and the rotation applied with them.

- **`AtAnyTiltTheBearingFollowsATurnAboutTheVertical`** holds a fixed tilt with
  both pitch and roll non-zero, turns the watch about the vertical, and
  requires the reported bearing to follow one for one.

Both were written because a tilt test about a **single** axis cannot tell a
correct projection from an incorrect one -- the two agree symbolically while
only one of pitch and roll is non-zero. That was verified rather than assumed:
with the projection deliberately replaced by a wrong-but-plausible composition
order, the single-axis test still passes and both of the above fail.

### Compatibility

`KERNEL_INTERFACE_VERSION` is not bumped. Nothing here changes an existing
message, layout or call; a new enum value and new headers are additive, and an
app built against the old SDK is unaffected. `SPO2` and `GRADE` were added the
same way.

`MAGNETIC_FIELD`'s field count goes from unspecified to four. The only code
that read the frame is the tutorial, by raw index rather than through a
parser, and nothing has ever produced a sample for it to read -- so there is
no consumer to break, and the tutorial is updated here anyway.

### Follow-up

The `una-kernel` PR follows, carrying the producer for both streams and the
submodule bump onto this.

## Summary by CodeRabbit

- **New Features**
  - Added support for corrected and raw magnetic-field sensor data, including X/Y/Z measurements and calibration status.
  - Added compass-bearing calculations for level and tilt-compensated readings.
  - Added validation for invalid, insufficient, uncalibrated, or unsuitable magnetic-field data.

- **Documentation**
  - Added guidance and examples covering magnetic-field parsing, calibration, bearing calculations, and compass integration.

- **Tests**
  - Added coverage for parsing, calibration, bearing calculations, tilt compensation, and invalid-data handling.


Docs/SensorsLayer.md
```diff
@@ -281,10 +345,10 @@ Blood-oxygen saturation derived from the optical PPG path. Delivered as a proces
 
 **Fields**: TOUCH (bool)
 
-For sensors without parsers (e.g. MAGNETIC_FIELD), use DataView:
+For sensors without parsers (e.g. HEART_BEAT), use DataView:
 ```cpp
 SDK::Sensor::DataView view = batch[0];
-float magX = view.f[0]; // Assume layout known from driver docs
+float value = view.f[0]; // Assume layout known from driver docs
```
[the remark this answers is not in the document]
sdvsaienko: Fixed in 0e8e2e25. None of the parserless sensors in the table documents any fields, so there is none to name. The example no longer names a sensor: it reads fields by index, using the layout the sensor's driver documents.

Libs/Header/SDK/SensorLayer/DataParsers/SensorDataParserMagneticField.hpp
```diff
@@ -0,0 +1,285 @@
...
+    float getAzimuthDeg() const
+    {
+        return isAzimuthValid() ? bearingDeg(getX(), getY()) : 0.0f;
+    }
+
+    /**
+     * @brief   Bearing clockwise from magnetic north, corrected for how the
+     *          watch is being held.
+     * @param   ax, ay, az: Gravity in the watch's axes, as the accelerometer
+     *          reports it. Any unit: only the direction is used. Take it from
```
rryles: A caller who reads "gravity" as *the vector pointing down* rather than as the accelerometer's reading will pass `(0, 0, -1)` here, and the result is silently wrong. It is also a mirror rather than a 180 deg flip, so it does not look wrong. For a field of `(10, 20, -40)`:

| gravity passed | bearing |
|---|---|
| level `getAzimuthDeg()` | 333.4 deg |
| `(0, 0, +1)` | 333.4 deg |
| `(0, 0, -1)` | 206.6 deg |

Nothing here can catch it, because face-down is a legitimate attitude - so the contract is the only defence. One clause would close it: something like "the accelerometer's own reading: a watch lying face-up at rest reads `az` positive".

Libs/Header/SDK/SensorLayer/DataParsers/SensorDataParserMagneticField.hpp
```diff
@@ -0,0 +1,285 @@
...
+public:
+    enum Field : uint8_t {
+        MAG_X = 0,  ///< Magnetic field along X, microtesla.
+        MAG_Y = 1,  ///< Magnetic field along Y, microtesla.
+        MAG_Z = 2,  ///< Magnetic field along Z, microtesla.
+        MAG_CALIBRATED = 3,  ///< 1 when a correction was applied to this sample.
+        COUNT
+    };
+
+    explicit MagneticField(const SDK::Sensor::DataView data) : mData(data) {}
```
rryles, line 60: `explicit` here, where every sibling parser's constructor is implicit (`Accelerometer(const SDK::Sensor::DataView data)` and the rest). So `MagneticField p = batch[i];` will not compile where it does for every other type.

Explicit is the better choice, so this is only worth a look in case the break from the pattern was not deliberate.

Libs/Header/SDK/SensorLayer/DataParsers/SensorDataParserMagneticField.hpp
```diff
@@ -0,0 +1,285 @@
...
+    }
+
+    /// Below this the horizontal field is too small to give a direction. Well
+    /// under any real horizontal field, which even at high latitude does not
+    /// come near it.
+    static constexpr float MIN_HORIZONTAL_UT = 3.0f;
+
+    /**
+     * @brief   Whether a field has a horizontal component worth a bearing.
+     * @note    Static and free of the sample, so the arithmetic can be
```
rryles: These three statics, plus `MIN_HORIZONTAL_UT` and `MIN_COS_PITCH_SQ`, are public and the note says the reason is that the tests can then reach them. On a repo whose readers are mostly outside the kernel team, that makes them API you cannot retract later.

`levelProject()` in particular is useful on its own - anyone doing their own fusion wants exactly it - so committing to them as supported and documenting them that way seems better than leaving them public for the tests' convenience. Making them private and testing through a constructed view is the other clean answer. Either is fine; it is the middle that costs later.

Tests/Host/sensorlayer/MagneticFieldParser_test.cpp
```diff
@@ -0,0 +1,365 @@
...
+    {
+        float xh = 0.0f;
+        float yh = 0.0f;
+        ASSERT_TRUE(MagneticField::levelProject(0.0f, northUt, -downUt,
+                                                0.0f, 0.0f, 1.0f, xh, yh));
+        level = MagneticField::bearingDeg(xh, yh);
+    }
+
+    // Pitch the watch nose-down by an angle: both the field and gravity rotate
+    // in the watch's axes by the same rotation about X.
```
rryles: This rotates about X, which in this file's own convention - pitch about Y, roll about X, as `levelProject()` takes them out of gravity - is a **roll**, not a pitch.

The test itself is right: `ay = -s, az = c` is exactly a roll and the assertion holds. Only the words are the wrong way round, which is worth fixing in a PR that turns on keeping the two straight.

Tests/Host/sensorlayer/MagneticFieldParser_test.cpp
```diff
@@ -0,0 +1,365 @@
...
+        // Turning the watch one way moves the bearing the other.
+        const float expected = std::fmod((reference - static_cast<float>(deg))
+                                         + 720.0f, 360.0f);
+
+        EXPECT_NEAR(std::fmod(got + 360.0f, 360.0f), expected, 0.2f)
+                << "the bearing did not follow the turn at yaw " << deg;
+    }
+}
+
+TEST(MagneticFieldParser, GravityThatSaysNothingIsRefused)
```
rryles, line 376: Three cheap gaps while the scaffolding is already here:

- Nothing checks that `getAzimuthDegTilted()` leaves `degrees` untouched when it fails *inside* `levelProject()`. `AnUncalibratedSampleHasNoBearing` covers only the early return; passing a zero gravity vector to the tilted call would cover the other path.
- The `MIN_COS_PITCH_SQ` gate is only exercised fully edge-on, so nothing pins where it actually falls over.
- Nothing records the accelerometer sign convention. A case asserting that `az > 0` and `az < 0` give different answers would at least make that contract fail loudly if it is ever misread.

Docs/SensorsLayer.md
```diff
@@ -133,6 +134,69 @@ void processBatch(uint16_t handle, SDK::Sensor::DataBatch& batch) {
...
+}
+```
+
+`getAzimuthDeg()` assumes the watch is held roughly level. To compensate for
+how it is actually being held, pass gravity in - read it from the
+accelerometer wherever you already read it:
+
+```cpp
+float bearing = 0.0f;
+if (p.getAzimuthDegTilted(ax, ay, az, bearing)) {
```
rryles, line 184: `ax, ay, az` arrive out of nowhere here. Worth a line saying they are the ACCELEROMETER sample's X/Y/Z taken close in time to this one - and, given the sign trap noted on the parser header, that it is the accelerometer's own reading rather than a down-pointing gravity vector.

Separately, and pre-existing, so entirely your call whether it belongs in this PR: the ACCELEROMETER row and field table above give the unit as `g`, but the kernel publishes m/s2 (`kFullScale * kGravity / kRawInt16Scale` in `ImuSubSensorAccel.hpp`, applied at `ImuSubSensorAccel.cpp:131`). It changes nothing for the new API, which uses only the direction.

Docs/SensorsLayer.md
```diff
@@ -19,7 +19,8 @@ All available sensor types are defined in [`SDK::Sensor::Type`](../Libs/Header/S
 | IMU | ACCELEROMETER_RAW | 0x11 | Acceleration raw | Yes | X,Y,Z (int16 raw) - 3 |
 | IMU | GYROSCOPE | 0x20 | Angular rate (3-axis) | Yes | X,Y,Z (float) - 3 |
 | IMU | GYROSCOPE_RAW | 0x21 | Angular rate raw | Yes | X,Y,Z (int16 raw) - 3 |
-| IMU | MAGNETIC_FIELD | 0x30 | Magnetic field (3-axis) | No | - |
+| IMU | MAGNETIC_FIELD | 0x30 | Magnetic field; corrected only when CALIBRATED | Yes | X,Y,Z (float uT), CALIBRATED (u32) - 4 |
```
rryles: Row says `CALIBRATED (u32)`, where the field table below and the enum itself both say `MAG_CALIBRATED`.

rryles (commented):
Reviewed with the maths checked independently rather than through the tests.

- Re-derived `levelProject()`: it is the standard `Rz(y)*Ry(p)*Rx(r)` eCompass decomposition, and the angles it pulls out of gravity match the rotation it then applies with them. Over 20,000 random attitudes (yaw +/-180 deg, pitch/roll +/-57 deg, inclination +/-69 deg, |B| 25-65 uT) the recovered bearing tracks true yaw to 5.7e-14 deg.
- Checked the claim about the two tilt tests by making it fail: with the projection replaced by the `Rx` then `Ry` composition order, `TiltingTheWatchDoesNotMoveTheBearing` still passes while `ProjectingGravityItselfLeavesNothingHorizontal` and `AtAnyTiltTheBearingFollowsATurnAboutTheVertical` both fail. They earn their place.
- Host suite builds and passes (16/16 magnetic, ctest green), `check_msvs_sources.py` clean across all 21 projects, and the tutorial is indeed the only consumer of the old `MAGNETIC_FIELD` frame.

The comments below are all minor. The one worth acting on is the gravity sign contract on `getAzimuthDegTilted()`.

Libs/Header/SDK/SensorLayer/DataParsers/SensorDataParserMagneticField.hpp
sdvsaienko: Agreed. Fixed in c5a359ad: the contract now says this is the accelerometer's own reading, az positive face-up, and that the opposite sign mirrors the bearing. A test pins it, using your field.

Libs/Header/SDK/SensorLayer/DataParsers/SensorDataParserMagneticField.hpp
sdvsaienko: Deliberate. The newer parsers are already explicit (BatteryLevel, Pressure, Temperature, MotionDetect, ...), so `Pressure p = batch[i];` does not compile either. Direct initialisation, as the docs use it, is unaffected.

Libs/Header/SDK/SensorLayer/DataParsers/SensorDataParserMagneticField.hpp
sdvsaienko: Kept public, as supported API. Fixed in c5a359ad: the notes no longer give tests as the reason, and SensorsLayer.md documents the statics for callers doing their own fusion. Making them private would weaken ProjectingGravityItselfLeavesNothingHorizontal.

Tests/Host/sensorlayer/MagneticFieldParser_test.cpp
sdvsaienko: Right, it is a roll. The comment is fixed in c5a359ad.

Tests/Host/sensorlayer/MagneticFieldParser_test.cpp
sdvsaienko: All three added in c5a359ad: a tilted refusal with zero gravity, the edge-on gate tested at 0.04 and 0.02, and face-up vs face-down.

Docs/SensorsLayer.md
sdvsaienko: Added under the snippet in c5a359ad: these are the X/Y/Z of a nearby ACCELEROMETER sample, passed as reported, with Z positive face-up. The g vs m/s^2 mismatch is real, but it predates this PR and does not affect the new API, which uses only the direction. Out of scope here.

Docs/SensorsLayer.md
sdvsaienko: Fixed in c5a359ad.

sdvsaienko:
Thanks for checking the maths independently. Addressed in c5a359ad, except the explicit constructor (deliberate) and the accelerometer unit (out of scope).

Docs/SensorsLayer.md
sdvsaienko: Tracked in #314.

rryles (approved):

PR #305, opened 2026-09-05
fix(appsystem): stop nothrow new pulling the exception runtime into every app

tobymurray:
`-fno-exceptions` never reaches precompiled libstdc++, so every app has been carrying ~10 KB of exception runtime and ~1 KB startup heap allocation for a catch that can't fire. `system.cpp` already replaces the other allocation operators, omitting the `nothrow` ones is what let the link fall through.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved memory allocation compatibility for builds that do not use exception handling.
  * Prevented unnecessary exception-runtime and ARM unwinder dependencies during linking.

### DIFF
--- a/Libs/Source/AppSystem/system.cpp
+++ b/Libs/Source/AppSystem/system.cpp
@@ -30,6 +30,7 @@
 #include <errno.h>
 #include <cstdint>
 #include <cstddef>
+#include <new>
 #include <atomic>
 #include <cstdlib>
 #include <cassert>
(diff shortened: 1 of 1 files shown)


rryles:
The premise and the fix both check out — forwarding the `nothrow` forms is safe here precisely because `operator new(std::size_t)` in this file is `noexcept` and returns `_malloc_r(nullptr, size)`, so null on failure. The nothrow contract survives the forward, which is the thing that would otherwise quietly break callers that check for null.

One gap worth closing while this is open: **the over-aligned operators are not replaced either, and they are the other way the runtime gets back in.**

`system.cpp` has no `align_val_t` overload at all, so a C++17 over-aligned type routes to libstdc++'s version exactly as the `nothrow` forms did. Compiling an over-aligned allocation with this repo's own app flags (`una-app.cmake:119` — `-fno-exceptions -fno-rtti -fno-use-cxa-atexit`, Cortex-M33):

```cpp
struct alignas(16) Wide { float v[4]; };
Wide* a = new Wide;
Wide* b = new (std::nothrow) Wide;
Wide* c = new Wide[n];
```

```
$ arm-none-eabi-nm -uC align_probe.o
U operator delete(void*, unsigned int, std::align_val_t)
U operator new[](unsigned int, std::align_val_t)
U operator new(unsigned int, std::align_val_t)
U operator new(unsigned int, std::align_val_t, std::nothrow_t const&)
```

Four symbols, none of them defined in `system.cpp`, so all four resolve into libstdc++ — including the aligned `nothrow` form, which is the direct analogue of the one this PR fixes.

Not theoretical on this platform: the kernel-side heap instrumentation for UNAWatch/una-kernel#291 attributed live allocations in a RunLVGL session to `operator new(unsigned int, std::align_val_t)`, which is why that build had to wrap `_ZnwjSt11align_val_t` alongside `_Znwj`. An app pulling in a type with extended alignment — anything `alignas`-ed, or a container of one — takes the same route.

The note in the new doc comment is already honest that the saving is conditional (an app linking `std::string` keeps the runtime regardless). This is a second condition in the same family, and closing it is four more forwarders of the same shape. `operator delete(void*, std::align_val_t)` and the sized/array variants come along for the ride.

Unrelated and pre-existing, just noting it since it is adjacent: `operator new[](std::size_t)` at line 432 is the only one of the six without `noexcept`, and the new `operator new[](size, nothrow_t)` is `noexcept` and calls it. Harmless as written, since it cannot actually throw — but the asymmetry invites someone to correct the wrong side of it later.

tobymurray:
I've added the aligned versions in [4561b4c](https://github.com/UNAWatch/una-sdk/pull/305/commits/4561b4c28fb81208c62053a038cc3ed539f6a30e). 

I don't think it's as straightforward as calling `memalign` though. The SDK redirects `fee` to the app's heap but not `memalign`, which still allocates from the kernel's heap. The app's `free` can't release that memory and every aligned `delete` leaks. That was already happening through libstdc++'s aligned new. 1f0c2309 allocates from the app's heap instead. 

`new[]` is now `noexcept` as well, as of bc9439ec

PR #318, opened 2026-09-17
fix(apps): lap numbering, activity end time, and the stale end-of-activity record

rryles:
Three defects found while decoding `.fit` files recorded on 1.5.0-rc4 (a Run, a RunLVGL run and a Walk), all in the per-app `ActivityWriter` / `Service` copies.

## 1. Every lap claimed `message_index = 0`

`ActivityWriter::addLap()` wrote a literal `0` into the lap's `message_index` field in all six activity apps. `message_index` (FIT field 254) is what consumers key laps by, so a multi-lap file collapses onto a single index â€” a 16-lap recording decoded as sixteen laps all reporting index 0.

Fixed by writing `mLapCounter`, which is 0-based and incremented immediately after the write. Each file has a second `.u16(0)  // message_index` in `stop()`; that one belongs to the *session* message and correctly stays 0, since there is one session per file. Its neighbouring `num_laps` already read the same counter, so `max(message_index) + 1 == num_laps` by construction.

Only visible on multi-lap files, which is why single-lap activities have always looked correct.

## 2. Elapsed time included the stop flow

An activity cannot be stopped without pausing first. `TrackActionPresenter::activate()` calls `trackPause()` unconditionally on entering the stop menu, and the only route to the save is the hold-to-confirm screen reached from there. So **every** activity accrues a trailing pause that is UI time, never movement â€” and it was being counted as elapsed.

Measured on the three rc4 files:

| Activity | Timer stop | Saved at | Tail counted as elapsed |
|---|---|---|---|
| Running | +931 s | +954 s | 23 s |
| RunLVGL | +932 s | +972 s | 40 s |
| Walking | +486 s | +497 s | 11 s |

40 s is 4% of a 15-minute run. Nothing flagged it because the session and final lap stayed self-consistent (`timeStart + elapsed == timestamp`) â€” the value was consistently wrong rather than obviously broken.

The activity now ends at the instant of that pause, with the same tail trimmed off the elapsed span. `timeStart` and the timer (active) total are unchanged, so the invariant still holds. A pause taken **mid-activity and resumed** is a real break and still counts in elapsed; only a pause that was never resumed is trimmed. The kernel-initiated stop path (`COMMAND_APP_STOP`) does not pause on its own, so it trims nothing when the track is running; if it fires while the track happens to be paused - the user sitting in the stop menu, or a long mid-activity pause - the activity ends at that pause, which is the right outcome.

`MonotonicCounter` already latched the instant in `mPauseStartValue` but exposed no accessor; `getPauseStartValue()` adds one.

## 3. The end-of-activity record published stale motion data

`Service::stopTrack()` wrote one final record to publish the latest battery sample, stamped at save time. Because that is after the timer stopped, it carried whatever motion the sensors still reported â€” a near-standstill speed:

| Activity | Final record speed | As pace |
|---|---|---|
| Running | 0.133 m/s | 125:19 /km |
| RunLVGL | 0.072 m/s | 231:29 /km |
| Walking | 0.102 m/s | 163:24 /km |

A pace graph draws that as a terminal spike, and auto-scaling then squashes the real 4:39â€“8:33 range into a sliver.

With fix 2 in place the activity ends at the pause, so a record stamped at save time falls outside the session entirely. The record is therefore **removed** rather than re-stamped onto a second the last periodic record already occupies. Consequence: the end-of-activity battery level and voltage are no longer recorded. `HRMonitor` is untouched â€” it has no pause concept and writes no trailing record.

## Scope

| App | Lap index | Trim end | Drop stop record |
|---|---|---|---|
| Running | yes | yes | yes |
| Cycling | yes | yes | yes |
| Hiking | yes | yes | yes |
| Treadmill | yes | yes | yes |
| Workout | yes | yes | yes |
| RunLVGL | yes | yes | yes |
| HRMonitor | yes | n/a (no pause) | n/a (none written) |

## Testing

`una-sdk-host-tests` builds clean and passes.

- `RunningActivityWriter.LapsGetSequentialMessageIndex` writes three laps end to end and asserts indices 0/1/2 plus the session's index and lap count. Verified as a real regression test: it fails against the previous literal `0`.
- `MonotonicCounter.PauseStartValueBoundsTheTrimmedEnd` reproduces the measured 931 s / 954 s case and asserts the trim arithmetic and `timeStart + elapsed == timestamp`.
- `MonotonicCounter.MidActivityPauseStaysInElapsed` asserts a 60 s mid-activity break survives the trim while the 10 s stop tail does not.

Only Running's `ActivityWriter.cpp` is compiled into the host tests, so the other five writers' identical lap-index edit rides on being textual copies. The `Service` trim is not host-testable directly; the tests cover the counter arithmetic the Services now use.

## Notes for review

- `mLapCounter` is reset only in `start()` and is not touched by `discard()`, `stop()`, pause/resume, the lap-reset helpers or `recoverInterrupted()`, so numbering cannot desynchronise from `num_laps`.
- `message_index` reserves bits 12-14, so a counter at or above 4096 would spill. That is 68 h of one-minute laps and `num_laps` shares the same `uint16` ceiling, so it is left unguarded.
- **Verified on device** (kernel 1.5.0-rc4, RunLVGL built from this branch): a 4-lap auto-lap activity and a pause/resume + 2-manual-lap activity. Lap indices sequential, no record after the final timer stop, no stop-flow tail, and the mid-activity pause correctly kept in elapsed while excluded from the timer. Full figures in the comments below.
- Not addressed here: records carry no `distance` field, so consumers must derive distance from speed or the GPS track. That is a format decision rather than a defect.

## Summary by CodeRabbit

* **Bug Fixes**
  * Lap records now use sequential message indexes, improving activity file accuracy across supported workout types.
  * Stopping an activity while paused now records the correct end time and excludes confirmation delay from elapsed duration.
  * Final activity records no longer include stale sensor or battery data after stopping a session.

* **Tests**
  * Added coverage for sequential lap indexes, session lap counts, and elapsed-time calculations around pauses.


Libs/Header/SDK/Metrics/MonotonicCounter.hpp
```diff
@@ -134,6 +134,14 @@ class MonotonicCounter {
      */
     bool isPaused() const;
 
+    /**
+     * @brief Value at the moment pause() was called.
+     *
+     * Only meaningful while isPaused() is true: the active value stopped
+     * advancing at this point, while add() keeps feeding the total.
+     */
+    T getPauseStartValue() const;
```
sdvsaienko: The `nowUtc / endUtc / tailSec` block is now pasted into `saveLap()` and `stopTrack()` of five apps, and the two new tests in `MonotonicCounter_test.cpp` re-type the same four lines rather than calling anything the Services call. If a Service copy drifts, the tests still pass. RunLVGL being missed (see the review summary) is the same problem from the other side.

Suggestion: put the arithmetic in the counter, where the state lives, e.g.

```cpp
/** @brief Last value that counts as activity: the pause instant while paused, else current. */
T getEndValue() const { return mIsPaused ? mPauseStartValue : mLastValidValue; }

/** @brief Span accumulated in the total since an unresumed pause; T{} when not paused. */
T getTrailingPause() const { return mLastValidValue - getEndValue(); }
```

Then each call site is three short lines, the tests exercise the real code, and `getPauseStartValue()` with its "only meaningful while isPaused()" caveat need not be public.

Examples/Apps/Running/Software/Libs/Sources/Service.cpp
```diff
@@ -1018,10 +1018,21 @@ void Service::saveLap(float autoLapDistanceM)
     // Save lap to the FIT file
     ActivityWriter::LapData fitLap{};
 
-    fitLap.timestamp = mTimeCounter.getCurrent();
-    fitLap.timeStart = mTimeCounter.getCurrent() - mTimeCounter.getLapValueTotal();
+    // Every user stop pauses first -- the GUI pauses on entering the stop menu
+    // and the confirm needs a hold -- so the span between that pause and the save
+    // is UI time, not activity time. End at the pause instant and trim the same
+    // tail from the elapsed span. No-op for a mid-activity lap, which is never
+    // written while paused.
```
sdvsaienko: This is true today, but only because of the GUI: R2 is handled on the track screen, which is not reachable while paused. Nothing in the Service enforces it -- `handleEvent(ManualLap)` (line 526) and the intervals phase advance (lines 1557 and 1573) call `saveLap()` regardless of track state.

If a lap were ever saved while paused and the track then resumed, the lap would end at the pause instant, but `resetLap()` consumes the full lap total, so the next lap starts at "now". Result: a gap between consecutive laps, and the lap elapsed values no longer sum to the session elapsed (which keeps a resumed pause).

Not reachable now, so not blocking. A one-line guard in the ManualLap handler (`if (mTrackState != Track::State::ACTIVE) return;`), or at minimum saying here that the GUI is what guarantees it, would make this robust against the next GUI change.

Same in Cycling, Hiking, Treadmill and Workout.

Examples/Apps/Running/Software/Libs/Sources/Service.cpp
```diff
@@ -1100,10 +1111,9 @@ void Service::stopTrack(bool discard)
             saveLap();
         }
 
-        mBatterySoc.request();
-        mBatteryVoltage.request();
-        ActivityWriter::RecordData fitRecord = prepareRecordData();
-        mActivityWriter.addRecord(fitRecord);
+        // No final record: the activity ends at the pause instant below, so a
+        // record stamped at save time would fall outside the session. The
+        // battery sample it used to carry goes with it.
```
sdvsaienko, line 1116: Agreed that a record stamped at save time has no place in the session any more. But the end-of-activity battery level/voltage is the only data point for "how much did this activity cost", and it is now gone.

An alternative that keeps it: request the battery and write that record in `pauseTrack(true)`. It is stamped at the pause instant, which is exactly the session end when the pause turns out to be the final one, and the motion data at that instant is real rather than stale. If the pause is resumed it is just one more record.

If nobody consumes the battery developer fields this can be ignored, but it is worth a conscious decision rather than a side effect. Same in the other four apps.

Examples/Apps/Running/Software/Libs/Sources/Service.cpp
```diff
@@ -1100,10 +1111,9 @@ void Service::stopTrack(bool discard)
...
 
         buildPartialSummary();
```
sdvsaienko, line 1118: `buildPartialSummary()` still sets `mSummary.utc = mTimeCounter.getCurrent()` (line 1087), i.e. the save instant, while the FIT session now ends at the pause. The JSON summary and the FIT file for the same activity therefore disagree by the trimmed tail. The "Track stopped. UTC: ..." log further down has the same offset, which will be confusing when comparing logs to a decoded file.

Using the same end value there keeps them consistent (trivial with a `getEndValue()` helper on the counter). Non-blocking; same in the other four apps.

sdvsaienko (changes requested):
Nice find, and the write-up made this easy to review. I checked the arithmetic against `MonotonicCounter` and it holds: `pause()` latches `mLastValidValue` and `add()` keeps feeding the total while paused, so `nowUtc - getPauseStartValue()` is exactly the unresumed tail, `timeStart` does not move, and `timeStart + elapsed == timestamp` survives. A resumed mid-activity pause stays in elapsed, as intended. The lap index fix is also right: `mLapCounter` is only reset in `start()` and `num_laps` already reads it.

**Blocking: RunLVGL is not covered.** It is in the tree, it has all three defects, and it is the app with the worst measured tail (40 s) in your own table, but it is absent from the Scope table and from the diff. At the PR head it still has:

- `.u16(0)  // message_index` in `addLap()` -- `Examples/Apps/RunLVGL/Software/Libs/Sources/ActivityWriter.cpp:284`
- the trailing `mBatterySoc.request()` / `addRecord()` in `stopTrack()` -- `Examples/Apps/RunLVGL/Software/Libs/Sources/Service.cpp:1103`
- untrimmed `timestamp` / `elapsed` in both `saveLap()` and `stopTrack()`

Its GUI pauses on entering the stop screen exactly like the TouchGFX apps (`LVGL-GUI/gui/src/screens/TrackActionScreen.cpp:51`), so the tail is still there. Please apply all three fixes to RunLVGL and add it to the Scope table.

The inline comments are suggestions, none blocking: the trim is duplicated ten times and the tests cover a copy; "never written while paused" is a GUI guarantee, not a Service one; the JSON summary still ends at save time; the battery sample could move to the pause instant instead of being dropped.

Description nit: `COMMAND_APP_STOP` -> `stopTrack(false)` does go through the trim when the track happens to be paused at that moment (user sitting in the stop menu, or a long mid-activity pause when the kernel stops the app). Ending at the pause is a sensible outcome, so no code change -- just "is never paused and is unaffected" is not quite accurate.

Since this is not verified on device yet, a real multi-lap capture before merging would be good, ideally one on RunLVGL too.

Libs/Header/SDK/Metrics/MonotonicCounter.hpp
rryles: Done in 009f973d, as suggested. `getEndValue()` and `getTrailingPause()` are on the counter, `getPauseStartValue()` is gone from the public surface along with its "only meaningful while paused" caveat, and each call site is now:

```cpp
const std::time_t endUtc  = mTimeCounter.getEndValue();
const std::time_t tailSec = mTimeCounter.getTrailingPause();
```

The tests now call those instead of re-typing the arithmetic, plus a new `EndValueIsCurrentWhileRunning` asserting the helpers trim nothing while running and that a *resumed* pause leaves no tail.

You were right that this is the same failure mode as the RunLVGL miss, and it was: my app survey ran against a checkout whose SDK predates 3fa51db9, so RunLVGL was not in the tree I enumerated and the duplicated arithmetic gave nothing to catch it. Both fixed in the same commit.

Examples/Apps/Running/Software/Libs/Sources/Service.cpp
rryles: Taken as a comment rather than a guard, in 009f973d. The comment no longer states it as a Service property:

> A mid-activity lap is not paused, so both calls are no-ops there -- though only the GUI guarantees that: the ManualLap handler and the intervals phase advance do not check the state.

I stopped short of the guard because it would also sit in front of the intervals phase advance, and I have not traced whether that path can legitimately fire while paused. Changing when a lap gets written is a behaviour change I would rather not make on a PR about timestamps, for something not currently reachable. Happy to add it if you would prefer it here.

Your analysis of the consequence is worth keeping on the record: `resetLap()` consumes the full lap total, so a lap saved while paused and then resumed would leave a gap between consecutive laps and break the sum against session elapsed.

Examples/Apps/Running/Software/Libs/Sources/Service.cpp
rryles: Not settled yet, and you are right that it should be a conscious decision rather than a side effect.

For context on how it got here: dropping the record was chosen deliberately over re-stamping it at the trim instant, because that second is already occupied by the last periodic record and a duplicate timestamp seemed the worse of the two. Writing it in `pauseTrack(true)` is a better answer than either, and it did not come up -- stamped at the pause instant, with real rather than stale motion, and merely one extra record when the pause is resumed.

It reverses a product call rather than a technical one, and it adds a record-write path to six apps, so I have left it for Ross to decide rather than folding it in. Flagging it here so the thread stays open until it is answered.

Examples/Apps/Running/Software/Libs/Sources/Service.cpp
rryles: Fixed in 009f973d. `mSummary.utc` and the "Track stopped. UTC" log both use `getEndValue()` now, so the `.json` summary, the log line and the FIT session all end at the same instant for one activity.

Applied to all six apps. Treadmill logs distance rather than a UTC at that point, so only its summary changed.

rryles:
Thanks for the review -- the blocking item was a real miss, and the cause is worth recording.

**RunLVGL is now covered** (009f973d). All three fixes applied: `message_index` in `addLap()`, the trailing `stopTrack()` record dropped, and both `saveLap()` and `stopTrack()` trimmed. Added to the Scope table in the description.

Why it was missed: the app survey behind the original scope was run in a checkout whose SDK submodule is pinned to 6d2b652, which predates 3fa51db9 (the RunLVGL merge). RunLVGL was not in the tree I enumerated, and I never re-derived the app list inside the branch, which is based on `origin/main` where it does exist. Enumerating from the wrong tree, then trusting that list.

Also in that commit, from your inline comments:

- the `nowUtc/endUtc/tailSec` arithmetic moved onto `MonotonicCounter` as `getEndValue()` / `getTrailingPause()`, so the ten call sites are two lines each and the tests exercise the real code
- `mSummary.utc` and the "Track stopped" log now share the FIT session's end instant
- the `saveLap()` comment no longer claims the Service enforces what the GUI actually guarantees

**Still open:** the battery record. Writing it in `pauseTrack(true)` is a better answer than dropping it and I have said so on that thread, but it reverses a product call, so it is with Ross rather than folded in here.

**Description corrected** on `COMMAND_APP_STOP`: you are right that it trims when the track is already paused. It does not pause on its own, so it trims nothing mid-activity, but "never paused and is unaffected" was wrong.

Device verification next -- a multi-lap capture on RunLVGL, since that is both the newly covered app and the one with the worst measured tail.

rryles:
## Device verification

Two activities recorded on a watch running kernel 1.5.0-rc4, with RunLVGL built from this branch (`RunLVGL_1.0.0.uapp`, 417,852 bytes, sha256 `739B5D8F27DE3AD1BBA9B72DA7D8DDC8115C89E2DF1F98FC10CB7B0F30890F94`). RunLVGL chosen deliberately: it is the newly covered app and had the worst measured tail before the fix. FIT files decoded with `fitdecode`.

### Run 1 — 1-minute auto laps

216 records, 554.4 m, 4 laps.

| Check | Result |
|---|---|
| lap `message_index` | `[0, 1, 2, 3]` (was `[0, 0, 0, 0]`) |
| session `message_index` / `num_laps` | `0` / `4` |
| records after the final timer stop | 0 |
| `total_elapsed_time` / `total_timer_time` | 215 s / 215 s — no tail |
| `start_time + elapsed == timestamp` | holds, session and all four laps |

### Run 2 — pause/resume, then two manual laps

169 records, 472.3 m, 3 laps. This one exercises the discrimination between a real break and the stop flow.

Timer events: `start` -> `stop @ +46 s` (pause) -> `start @ +65 s` (resume) -> `stop @ +187 s` (final).

| Check | Result |
|---|---|
| lap `message_index` | `[0, 1, 2]` |
| records after the **final** timer stop | 0 |
| session end vs last record | 0 s apart — stop-flow tail gone |
| `total_elapsed_time` / `total_timer_time` | 188 s / 169 s |
| the 19 s mid-activity pause | **kept** in elapsed, excluded from the timer; visible as the single 20 s record gap |
| `start_time + elapsed == timestamp` | holds, session and all three laps |
| lap sums vs session | elapsed 188 == 188, timer 169 == 169 |

Lap 0 spans the pause (elapsed 118, timer 99); laps 1 and 2 contain no pause and have elapsed == timer. So only an unresumed pause is trimmed, which is the behaviour the tests assert.

### Unrelated observation from the same session

Worth flagging because it showed up while testing, but it is not from this branch — nothing here touches allocation.

After one RunLVGL activity, every subsequent large-GUI launch in the same boot fails:

```
Loader::load: App [RunLVGL.GUI] required RAM: 454076 bytes
sbrk: newlib heap exhausted, malloc will return NULL (incr=454144, headroom=401184)
Loader::load: Failed to allocate 454076 bytes for the app
Application::runGui: Failed to load GUI
```

The service (111,456 B) loads every time; only the GUI fails, and the headroom figure is identical (401,184) across roughly a dozen attempts. **TouchGFX Run fails the same way in the same session** (needs 492,724 B), so it is not LVGL-specific and not app-size-specific — the kernel heap has no single free block that large any more and the loader needs one contiguous chunk. Only a reboot clears it; the first launch after boot always succeeds. This is the same mechanism previously diagnosed behind the watch-face-gallery launch failures, which predicted exactly this for Run and Treadmill. Evidence in `D:\Log\System\syslog_709724_0047_20260917T150916.log`.

sdvsaienko:
Checked 009f973d against the code, not just the replies -- all good.

- **RunLVGL** has all three fixes; its `Service.cpp` is now byte-identical to Running's, so there is nothing app-specific left to drift.
- `getEndValue()` / `getTrailingPause()` on `MonotonicCounter` are exactly what I had in mind, `getPauseStartValue()` is gone from the tree, and every call site in all six apps is down to the two lines. The tests now go through the real helpers, and `EndValueIsCurrentWhileRunning` covers the case I cared most about (a resumed pause leaves no tail).
- `mSummary.utc` and the "Track stopped" log share the FIT end instant everywhere; Treadmill having only the summary change is correct.
- The reworded `saveLap()` comment is fine. Agreed that a guard in front of the intervals phase advance is a behaviour change that does not belong in a PR about timestamps -- the comment is enough for something unreachable.

The device run on RunLVGL is the evidence I was after, in particular run 2: the 19 s resumed pause kept in elapsed, the stop tail gone, and lap sums matching the session. Thanks for choosing the newly covered app for it.

Two leftovers, neither blocking:

- The battery record: happy for that to be its own PR once the product call is made. 
- Nit: `PauseStartValueBoundsTheTrimmedEnd` is named after an accessor that no longer exists; rename whenever that file is next touched.

sdvsaienko (approved):

Issue #300, opened 2026-08-29
Phone notifications render Hebrew as fallback `?` — notification font lacks the Hebrew block


noamsto:
Hebrew notifications forwarded from the phone display as runs of `?` in the notification overlay — no Hebrew glyphs are drawn.

## Why this looks like glyph coverage alone

`?` is TouchGFX's fallback character, documented in the vendored core as "the fallback character to use when no glyph exists for the wanted character" (`ThirdParty/touchgfx/framework/include/touchgfx/Font.hpp:570`). So the text is reaching the renderer as discrete, correctly-decoded characters and failing only at glyph lookup.

Two observations narrow it further:

- The number of `?` per word tracks the Hebrew letter count rather than doubling it. Hebrew is two bytes per letter in UTF-8, so a byte-level decoding bug would produce roughly twice as many substitutions.
- Spaces are preserved as spaces. U+0020 has a glyph; the Hebrew letters do not.

Together those suggest the UTF-8 to UTF-16 conversion in the firmware is correct and the notification pipeline is sound end to end. The missing piece appears to be the glyph data in the notification font.

## The gap looks narrow

Hebrew is a cheaper script to add than most:

- **27 codepoints**, contiguous at U+05D0–U+05EA, including the five final forms. Unpointed Hebrew — which is what notification text is — needs no niqqud.
- **No contextual shaping.** Arabic needs per-position letter forms, which is why the core carries `adjustArabicGlyph()`. Hebrew needs none of that machinery.
- **RTL is already compiled in.** `drawStringRTL()` is present in the vendored TouchGFX 4.26.1 core (`framework/include/touchgfx/lcd/LCD.hpp:798`), documented for "sequences of Arabic / Hebrew and Latin", and `TEXT_DIRECTION_RTL` is defined in `hal/Types.hpp:605`.

So the work looks like adding Hebrew glyph coverage to the notification font at the sizes already in use, plus setting RTL base direction on the notification text area. Without the direction flag the glyphs would render but in reversed word order.

## Ask

Is Hebrew notification support something you would consider? Happy to test a build against real hardware and report back.

Worth noting the SDK gives apps no way to work around this: `IAppCapabilities::enablePhoneNotification(bool)` is a suppress toggle, and no message in `Libs/Header/SDK/Messages/` carries notification content, so an app cannot re-render the text itself. It has to be solved in firmware.

noamsto:
Observed on firmware 1.4.0 (latest at the time of filing), iPhone.

PR #319, opened 2026-09-21
fix(port): return queued custom messages before the GUI exits

rryles:
## The bug

`GuiCommandProcessor` parks application-specific messages in its own `mUserQueue` for the app's code to collect later. On `COMMAND_APP_STOP` it called `exit(0)` with messages still sitting there.

Those are the kernel's pool blocks. The kernel hands the pointer over and gets it back only when the app releases it, and no kernel-side queue drain can see this queue — so anything parked here when the process exits is never returned. Measured in the field: rapid app cycling exhausted a 100-block message pool in under a minute, and it did not recover for the rest of the boot.

## The fix

Drain `mUserQueue` on stop, failing each queued message back before `onStop()` and `exit()`:

- each message gets `FAIL`, then `sendResponse`, then `releaseMessage` — the same sequence the queue-full eviction path and the normal handled path already use;
- the count is logged only when non-zero, so an ordinary stop stays quiet;
- a message that could not be pushed into `mUserQueue` is now answered rather than released silently by the shared tail, so a sender blocked on it is not left to time out.

`sendResponse` before `releaseMessage` is required, not cosmetic: `releaseMessage` can drop the last reference, and at zero the kernel deletes the completion semaphore. Responding afterwards would signal a deleted semaphore.

## Scope, and what this does not do

This is the cooperative half. Kernels that reclaim a dead process's blocks recover these anyway, so where that exists this costs latency rather than the block itself; on kernels without it, the block is lost for the boot.

The present-day value is simply returning the block, which matters on kernels without that reclaim.

There is a second thing the drain does that a reclaim cannot — it answers, where the reclaim drops the reference silently, so a sender blocked in `waitCompletion` would otherwise wait out its own timeout. That is a correct invariant but **not currently reachable**, and it is worth being precise rather than claiming it as live value: `mNeedsResponse` is set only by `initCompletionSemaphore`, which runs only when `timeoutMs > 0`, and both `sendToService` and `sendToGui` hardcode `timeoutMs = 0` to avoid deadlock. So the kernel never hands an app a message that wants a response. The only route to one is an app-specific `sendMessage` with a timeout between service and GUI, and no such call exists in the tree. `sendResponse` in the drain is therefore a no-op for everything that can be parked there today — correct defensive code that a third-party app could reach tomorrow, not a scenario running now.

Two gaps remain open deliberately, because closing either is a wider change than this one:

- **App-initiated exit.** Apps that quit via `Model::exitApp()` → `sys.exit()` never reach `COMMAND_APP_STOP`, so the drain does not run. Closing that means exposing the drain and routing app exit through it.
- **A forced unload**, where the GUI never processes the stop at all, cannot be handled from inside the app by construction.

## Note for reviewers

`mUserQueue` is a `FixedQueue` with no locking, and `mSize` is a plain `size_t` incremented by the pump thread and decremented by the handler thread. On the watch that is safe — both run on the same thread. In the TouchGFX simulator they are different threads, so the race is already present at one push and one pop; this drain widens it to up to ten pops at shutdown rather than introducing it. Simulator-only and not addressed here — tracked separately, since it outlives this PR.

## Testing

Compiles clean for Cortex-M33 with the production flag set. The behaviour is exercised on hardware through the kernel-side counterpart's test: rapid app cycling over 97 seconds, 313 process teardowns, zero pool allocation failures.

## Two related fixes carried in the same PR

Both came out of reviewing the kernel counterpart, and both are about the same
rule: **a message type must not declare a destructor.**

The kernel destroys messages non-virtually, because a message an app allocates
is constructed inside that app's image and the kernel frees that image when the
process unloads -- dispatching through the vtable there would jump into freed
memory. So a destructor declared on a message type silently does not run.

**`MessageBase.hpp` now says so.** It previously said the opposite ("virtual
destructor for proper cleanup of derived classes"), in the one header every app
includes. It cannot be a `static_assert`: the virtual destructor makes every
derived type non-trivially-destructible, so the trait that would express the
rule is false for all of them. The destructor stays virtual -- the vptr is at
offset 0 of the published layout and removing it would change the size and
shape of every message across the ABI.

**The simulator now destroys messages the same way the watch does.** It called
`delete msg`, which dispatches the virtual destructor, so an app that broke the
rule would see its destructor run in the simulator and not on the device -- the
simulator would hide the very bug it exists to surface.

That same line also leaked: `allocateRawMemory` takes the storage with
`new uint8_t[size]`, but `MessageBase::operator delete` is a no-op, so
`delete msg` returned nothing. Every message the simulator allocated was
leaked. The storage is now freed with the `delete[]` that matches how it was
taken.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved application shutdown handling so pending commands receive clear failure responses instead of remaining unresolved.
  - Improved error handling when commands cannot be queued, ensuring affected requests are completed with an appropriate failure status.
  - Improved message cleanup during shutdown and command processing to reduce stalled or incomplete requests.
  - Improved reliability when handling message allocation failures, helping prevent unexpected behavior during command processing.


sdvsaienko:
**Approve.** The drain is right, the ordering argument for `sendResponse` before
`releaseMessage` is right, and answering the unreachable `!messageQueued` branch
rather than leaving the one silent drop in the function is the right call.

One correction to the description rather than to the code.

The "Scope" section rests the PR's independent value on this:

> the sweep drops the reference without answering, so a sender blocked in
> `waitCompletion` is freed only by its own timeout, whereas the drain here
> wakes it immediately

That is a correct invariant but not currently a reachable one, and the reason is
one level down: `mNeedsResponse` is set only by `initCompletionSemaphore`, which
runs only when `timeoutMs > 0` — and `DualAppComm::sendToService` / `sendToGui`
both hardcode `timeoutMs = 0` ("avoid deadlock"). So the kernel never hands an
app a message that wants a response. The only way one can reach `mUserQueue` is
an app-specific `sendMessage` with a timeout between service and GUI, and there
is no such call anywhere in the tree: the only two-argument `sendMessage` calls
are `GuiCommandProcessor.cpp:228` (`RequestDisplayUpdate`, 1000) and the FitFiles
tutorial's `RequestGlanceUpdate` / `RequestGlanceConfig` (100), all three
kernel-bound, i.e. the waiter is the sending process itself.

So `sendResponse` in the drain is a no-op for everything that can actually be
parked there today. That does not make it wrong — it is correct defensive code
and a third-party app could reach it tomorrow — but the paragraph currently reads
as a live scenario, and the drain's real present-day value is simpler: returning
the block, which matters on kernels without #300.

Two small things:

- Your "Note for reviewers" undersells the simulator race slightly.
  `FixedQueue::mSize` is a plain `size_t` incremented by the pump thread and
  decremented by the handler thread, so the UB is already there at one push and
  one pop — the drain widens it rather than introducing a new kind. Your proposed
  follow-up (guard the queue, or drive the pump from one thread in the simulator)
  is the right fix; worth a tracked issue rather than a PR note, since it
  outlives this PR.
- Same file as your third commit: the simulator's `allocateRawMemory` does not
  `memset` while the kernel's does. Same category of sim↔watch divergence the
  commit exists to close — a message field the constructor does not initialise
  reads as garbage in the simulator and zero on the device. One line, your call
  whether it belongs here or separately.

Optional: `Docs/TouchGFX-Port-Architecture.md` is where an app author actually
looks, and it already documents the queue-full eviction. Two things from this PR
belong there — that pending custom messages are failed and released on stop, and
the no-destructor rule (which currently lives in a `protected:` section of
`MessageBase.hpp`).

sdvsaienko (dismissed):

rryles:
Thanks — the reachability correction is the important one and you are right. I had written the blocked-sender wake-up as live value when it is a correct-but-unreachable invariant: `sendToService`/`sendToGui` both hardcode `timeoutMs = 0`, so `initCompletionSemaphore` never runs for a kernel→app message, and there is no app-specific `sendMessage` with a timeout in the tree. Description now says the present-day value is returning the block, and that `sendResponse` is defensive code a third-party app could reach rather than something happening today.

Race note corrected too — `mSize` being incremented and decremented by different threads means the UB is there at one push and one pop, so the drain widens it rather than introducing it. Filed as #320 rather than left as a PR note, with both fix shapes: guard the queue, or drive the pump from the SDL thread so the simulator matches the watch's single-consumer model. I lean toward the second as the fidelity fix, but it touches simulator startup for every example app.

`allocateRawMemory` now memsets — 76512dfd. Same argument as the destructor commit: an app should not be able to pass in the simulator and misbehave on the device for a reason that has nothing to do with the app.

On the docs: agreed `Docs/TouchGFX-Port-Architecture.md` is where an app author actually looks, and both items belong there — the stop drain alongside the queue-full eviction it already documents, and the no-destructor rule, which is currently buried in a `protected:` section of `MessageBase.hpp`. I have left that out of this PR to keep it to the fix; happy to do it here if you would rather it land together than as a follow-up.

sdvsaienko:
`76512dfd` looks right, including the `return ptr` you added on the failure path
so the memset is not reached with a null — that was not in what I flagged.
Filing the `FixedQueue` race as #320 with both fix shapes is better than the PR
note, and I agree the single-consumer simulator is the fidelity fix rather than
the lock; the cost you name (simulator startup for every example app) is real
though, so the guard may be the thing that actually ships first.

On the docs: **separate PR, please.** Three reasons, none of them about the
content, which I think is right:

- This PR is a tight three-commit fix that has already been read at this shape.
  A docs commit gives it a second review surface for no gain in the fix.
- A new or substantially changed page under `Docs/` drags in `Docs/index.rst`
  and the built public site, which is its own check rather than a rider on a
  port fix.
- The scope is bigger than the two lines this PR would justify. The no-destructor
  rule is not really a TouchGFX-port fact — it belongs wherever an app author is
  told how to define a message, so the right change is probably
  `TouchGFX-Port-Architecture.md` (stop drain, next to the queue-full eviction it
  already documents) *and* the message/SDK overview path, decided together
  rather than piecemeal.

Happy to review that one when it lands.

sdvsaienko (approved):

rryles:
Agreed on all three counts, and the third is the one that changes what I would have written: the no-destructor rule is not a TouchGFX-port fact. I was about to put it in `TouchGFX-Port-Architecture.md` because that is where the queue-full eviction already lives, which would have filed it under the port rather than under how you define a message — findable only by someone already reading about the port, which is not who needs it.

So: separate PR, and scoped as you describe — the stop drain next to the eviction it belongs beside, and the no-destructor rule wherever an app author is told how to declare a message type, decided together rather than one line in each.

This PR stays as it is. Nothing has been pushed to it since your approval, so it is still green and approved at 76512dfd.

Noted on #320 too — you may well be right that the guard ships first. The simulator-startup change touches every example app's `main.cpp`, so it is the correct fix carrying the higher cost, which is usually how these go. I left both shapes in the issue rather than picking one for exactly that reason.

